PR #53, opened 2014-04-27
Update res_company_view.xml

betavisionsys:


rvalyi:
obrigado @JoaoPauloSouza !

PR #128, opened 2015-05-28
_fiscal_comment for Simples Nacional

pauloneda:
```
def _fiscal_comment(self, cr, uid, order, context=None):

    if order.company_id.fiscal_type == '1':
        calc = cur_obj.round(cr, uid, cur,(order.amount_total * (order.company_id.taxa / 100)))
        fp_comment = fp_comment[0].replace('R$','R$ '+str(calc))

    return fp_comment
```


rvalyi:
@pauloneda this PR looks weird. There shouldn't be as may commits. You probably picked the wrong target branch or made some other error.

rvalyi:
@pauloneda esse PR foi fechado automaticamente ao fazer https://github.com/OCA/l10n-brazil/issues/256
Se ainda for relevante por favor faca de novo o PR na nova branch 7.0 (antiga develop).

PR #124, opened 2015-02-12
8.0 branch

mileo:
Hello,

A few months ago I created a "PR" for the creation of a branch 8.0 for this repository. 

> Can someone that have push permission to upstream, create the 8.0 branch? https://github.com/OCA/l10n-brazil/pull/109

Now this is a pull request of months of work from @danimaribeiro @mileo @mstuttgart @Mackilem to migrate l10n_br to 8.0. Here is the currenty WIP:
1. Make all modules installable; OK
2. Fix bugs and make then operational; OK
3. Migrate to new API and flake8 pass:
- [X] l10n_br_account
- [ ] l10n_br_account_payment
- [x] l10n_br_account_product
- [ ] l10n_br_account_product_service
- [ ] l10n_br_account_service
- [ ] l10n_br_account_voucher
- [x] l10n_br_base
- [x] l10n_br_crm
- [ ] l10n_br_crm_zip
- [x] l10n_br_data_account
- [x] l10n_br_data_account_product
- [ ] l10n_br_data_account_service
- [x] l10n_br_data_base
- [x] l10n_br_delivery
- [ ] l10n_br_hr_timesheet_invoice
- [x] l10n_br_product
- [x] l10n_br_purchase
- [x] l10n_br_sale
- [x] l10n_br_sale_product
- [ ] l10n_br_sale_service
- [x] l10n_br_sale_stock
- [x] l10n_br_stock
- [ ] l10n_br_zip

During migration we also did everything possible to improve functionality. There have been improvements in crm ( company and contact creating on lead ) and account ( full operational MD-E ). If anyone wants to help or is already working on migration we will be happy to know.

In https://github.com/OCA/l10n-brazil/pull/109#issuecomment-64740403 and as told before yesterday on skype (with @danimaribeiro): @renatonlima informed us that he had worked on migration in November and publish your code today. For some unknown reason the branch 8.0 has not already been created and we had no feed back. So that's the PR of our work.

But both KMEE as other contributors are waiting a long time to be able to contribute more actively and without a very large delay. We would like to request that someone looked fondly our actions so that we can continue to contribute more and better. Thus, we would not have only one company handling the official repository of the Brazilian location. We can contribute to reviewing and analyzing third-party code task to be able to streamline the approvals in our location.

[...]

PR #130, opened 2015-06-08
Migration/l10n br crm

renatonlima:


mileo:
@renatonlima are you redoing the work of https://github.com/OCA/l10n-brazil/pull/124 ?

rvalyi:
so once this branch with l10n_br_base https://github.com/OCA/l10n-brazil/pull/129 is passing tests again, it needs to be merged into this one to enable the installation and testing by Travis.

rvalyi:
So I didn't merge l10n_br base in this branch to avoid possible conflicts when merging globally. Instead here is the test branch where we also have code-metrics: https://github.com/OCA/l10n-brazil/pull/154

The coverage is lower with the l10n_br_crm module and I think it deserved a dumb lead conversion test before we merge.
Lint results are here: https://travis-ci.org/OCA/l10n-brazil/jobs/66147684
We need them green before considering merge, so :-1: for now.

@mileo. Same as for the l10n_br_base PR. Please if we are missing any good feature you guys did, please cherry pick the commits and rebase them on this branch We will do our best to look at your work before merging anyway. Sorry again for the inconvenience of the double work here, again explained by the very exceptional conditions we had.

l10n_br_crm/models/__init__.py
```diff
@@ -1,7 +1,7 @@
-# -*- encoding: utf-8 -*-
+# -*- coding: utf-8 -*-
 ###############################################################################
 #                                                                             #
-# Copyright (C) 2011  Renato Lima - Akretion                                  #
+# Copyright (C) 2015  Renato Lima - Akretion                                  #
```
rvalyi: should be 2011 - TODAY

rvalyi:
:+1: 
PEP is OK, installation works on the test branch.
We can eventually add a lead conversion test.

rvalyi:
@mileo I see that in your branch you added the rg on a lead (seems legitimate), you also separated the cpf and cnpj fields. Could you elaborate on why (it's not in the commit message I think)? @renatonlima what do you think about this design? https://github.com/odoo-brazil/l10n-brazil/blob/8.0/l10n_br_crm/crm_lead.py#L42 cc @danimaribeiro

rvalyi:
OK, so we are now green on tests and PEP-8.
That's sad to decrease the coverage, so that would be cool to add some basic lead conversion test and keep a high coverage.

@mileo we found some design changes you did here like separating the cpf and cnpj fields. Could you please argue why you think it's better than what we have now? @renatonlima didn't seem convinced but may be you'll convince us.

danimaribeiro:
One is used for the company data, and the other for the contact. 
In this case, I think is necessary to have both fields.

capturar

Not ok to merge yet.

danimaribeiro:
It happens an error when converting a lead to an oportunity.

error creating lead

The method signature is different from the crm base class. Adjusting the signature to the following:

new signature

It works partially, because now I get the error about "CNPJ invalido":
cnpj invalido

So definitely we need two different fields, CNPJ and CPF, one is for legal_name and the other for contact_name.

danimaribeiro:
It's better to get the code from odoo-brazil, with @api.model, the lead conversion to opportunity is working nicely, generating both company and contact with the corresponding CNPJ and CPF.

It would be nice to have the address fields order according to partner and company, but that I can submit a patch after.

danimaribeiro:
Merged in #165

rvalyi:
oups that one had to be closed then

PR #143, opened 2015-06-09
Migration/l10n br product

renatonlima:


rvalyi:
:-1: 

Sorry but this module is now empty now that the logic moved to l10n_br_account_product. The only thing it does is setting the pricelist in R$ money. We can have that in other modules. I don't think we need this module anymore just for this setting.

rvalyi:
So I moved the pricelist setting into the the l10n_br_account module to avoid this almost empty module.

mileo:
:1:

mileo:
:+1:

PR #134, opened 2015-06-09
Migration/l10n br account product

renatonlima:

PR #144, opened 2015-06-09
Migration/l10n br purchase

renatonlima:


rvalyi:
will need a rebase once l10n_br_stock is merged. But we made tests locally with a local l10n_br_stock copy meanwhile and fixed the code so reviews can already start.

rvalyi:
rebased on 8.0, that is including the l10n_br_stock dependency. However we got again a strange "heisenbug" in the l10n_br_account module that seems totally unrelated to this PR https://travis-ci.org/OCA/l10n-brazil/jobs/92494378#L786
@renatonlima this one is driving us crazy...

mileo:
This is the error that @lfdivino had 9 days ago

https://github.com/OCA/l10n-brazil/pull/134#issuecomment-156265177

rvalyi:
@mileo @lfdivino yes, we never had it at the time he reported it and couldn't find any explanations for it. But then we had it a couple a of times without being able to find any explanation. Seems same code will often pass and fail in rare exceptions. Any help for finding the cause of the bug or a workaround is welcome. I wonder if some database lock may cause a workflow transaction to fail inconsistently or something like that...

renatonlima:
@rvalyi @mileo @lfdivino @mbcosta

I found the error and the problem is in travis all modules are installed and after run all tests for all modules in branch, and now we have a conflict in account.invoice workflow, at moment I change all tests in l10n_br_account to do not confirm any invoice until migration is finish I made a PR to solve this https://github.com/OCA/l10n-brazil/pull/242

rvalyi:
Travis green; @mileo @danimaribeiro reviews welcome...

mileo:
@kmee ! Review please

rvalyi:
:+1: LGTM

mileo:
:+1:

l10n_br_purchase/__openerp__.py
```diff
@@ -1,44 +1,44 @@
...
+        'views/res_company_view.xml',
         'security/ir.model.access.csv',
         'security/l10n_br_purchase_security.xml',
     ],
     'demo': [
-        'test/purchase_order_demo.yml'
-            ],
-    'installable': False,
+        # FIXME
+        # 'test/purchase_order_demo.yml'
```
mileo, line 40: Is this right?

renatonlima:
Yes @mileo, at moment there is not test on purchase, I working in some tests in sale modules, after I will complete this test in purchase.

mileo:
Sale module we cant make tests pass here. =(
How is the way to do it with web_context_tunnel?

renatonlima:
You can define context values in yml like this https://github.com/OCA/l10n-brazil/blob/8.0/l10n_br_purchase/test/purchase_order_demo.yml#L5

mileo:
We are testing the https://github.com/OCA/l10n-brazil/pull/148 with the same approach but does not works well.

PR #150, opened 2015-06-09
Migration/l10n br zip

renatonlima:


danimaribeiro:
File res_partner_view.xml should not set width 50% on zip, it's proposed to do in base:
https://github.com/OCA/l10n-brazil/pull/160

The following error is happening when we enter the partner view:
AttributeError: 'res.partner' object has no attribute 'onchange_mask_zip'
because the res_partner does not implement onchange_mask_zip.

danimaribeiro:
View of CEP needs some polish.
tela-cep

danimaribeiro:
Zip search is throwing the exception:
"ValueError: dictionary update sequence element #0 has length 1; 2 is required" when calls .map_to_zip_result()

danimaribeiro:
:+1:

PR #152, opened 2015-06-09
Migration/l10n br data base

renatonlima:


rvalyi:
:+1: this is only a data module. Lint tests are passing. The installation test fails because of the missing l10n_br_base module in this branch, but it will be found again once merged on 8.0 if we merge it after l10n_br_base.

rvalyi:
I'm also merging that one as it contains no Python code and it the version from @danimaribeiro and @mileo didn't touch that module https://github.com/odoo-brazil/l10n-brazil/tree/8.0/l10n_br_data_base

danimaribeiro:
There are a few data files in l10n_br_base, especifically city and state, should we move those to here?

PR #154, opened 2015-06-10
Migration/l10n br crm tests

rvalyi:
Branch to explicit tests and code-metrics. DO NOT MERGE
Merge migration/l10n_br_crm instead.


rvalyi:
closing as the standard crm migration feature is now passing now that we could rebase on 8.0 including the base modules.

PR #157, opened 2015-06-18
Remigration/l10n br base

mileo:


danimaribeiro:
I'll fix the build.

rvalyi:
guys, in this cases, I explained we found no added feature vs the previous Akretion work. So as we had a better coverage and code layout and as we are the original authors of the module I merged it. This also makes it easier to get Travis run the tests in the subsequent modules such as the CRM where we can debate a small design change you did BTW.

Overall, I'm not sure it's really useful you try to split your branch because when we check for features, what we look here are commit messages to spot features and files and we are able to do that even with your current branch layout.

I think something more useful would be you help us spot these features you added over the PEP-8 and API migration which should roughly be the same as it's quite automatic. If you think you really did something smart in some API refactoring or if we we forgot some API refactor that you did already by luck, we will also take it.

Also, for these extra features you coded, like the MDE fiscal invoices, that would be cool if you could create some test. May be we won't require it because we don't have an exhaustive coverage anyway, but if you have some, that would be awesome.

Again, we know very well the situation is annoying. I hope you guys understand we had no choice unfortunately cause we would have broken if we had triggered the v8 market without being able to serve it for our own critical customers anymore. But what is important now is to move on and focus on the code quality and the real features.

rvalyi:
also, as for cherrypicking the feature commits. I think it will mean manually applying the code change and the commit with the proper author. Indeed, as history and code layout differ, it will hardly be automatic. Eventually if a feature you did is really isolated in a set of commits, then if you change the code layout, may be a patch can be applied automatically, but I would say it's probably not worth the trouble.

Also guys, you might be angry because for the 1st time in 6 years we had no community activity in 6 months. But guys this is not the only project where it happens and sadly that was "force majeure" in this case. So for instance I'm really fan of a CMS called LocomtiveCMS. Rails 4 has been around for 2 years already and LocomotiveCMS doesn't support it yet. Yes, there are a few folks that tried to get LocomotiveCMS running on Rails 4 and submitted some branches around. But none of them came angry to the project author and tried to highjack the project leadership to get Locomotive 4 use their own Rails 4 branches just because at some point their threw it earlier in the air... You see what I mean?

rvalyi:
closed because superseded by https://github.com/OCA/l10n-brazil/pull/129
meanwhile all changes to l10n_br_base were carefully checked and what was missing was cherry picked. Feel free to send extra PR if you spot something we missed.

PR #165, opened 2015-07-30
Migrate/crm to Odoo 8.0

danimaribeiro:
New PR with the changes from odoo-brazil.
It's necessary to create those new fields (cpf, cnpj, rg)


renatonlima:
Hi Danimar,

I aggre with new field RG, but I not sure about split cnpj and cpf in tow fields, I will review this code before merge this PR

danimaribeiro:
When you fill all the fields, partner and contact in the lead view, Odoo calls _lead_create_partner twice, the first time it passes is_company=True, and is_company=False, creating two partners, it's not right to use the same cnpj_cpf field for both created partners.
If you want to create a lead for "pessoa fisica", you just put fill the fields contact and cpf, when you create the partner, it will create just one partner with is_company=False and the correct cpf field.

mileo:
In the end cnpj goes to Company (is_company = True)  cnpj_cpf and cpf goes to Contact(is_company = False)

PR #142, opened 2015-06-09
Migration/l10n br hr timesheet invoice

renatonlima:

PR #173, opened 2015-08-24
Migration/l10n br timesheet

danimaribeiro:


rvalyi:
Hello @danimaribeiro sounds cool. Now as this is contributed code from Kmee we would be more confortable if there could be a test at least passing in the invoice_cost_create method. In this 8.0 OCA branch we are indeed trying to keep a high test coverage, this also to make future migrations less painful. Also to make tests and installation pass, that would be cool if you could rebase on the last 8.0 branch. 
Thanks for the work at this point anyway.

danimaribeiro:
I don't know exactly what is used for, I just fixed pep. I think Kmee could provide a better test to it.
I don't think is necessary to rebase, I just improve the code from this:https://github.com/OCA/l10n-brazil/pull/142 and the travis build is passing.

rvalyi:
ping @mileo we would happily merge that but that would be cool if you could implement a test so we don't destroy the coverage and get ready for v9 migration...

rvalyi:
@mileo if you participate to a codesprint on l10n-brazil tomorrow, that would be cool if some tests could be added here so we can merge it.

mileo:
I think that i will work in cnab

rvalyi:
@mileo I understand you are looking for a fast solution. But really in just a few days of work I could plug brcobranca and support many banks in a well tested and well supported manner, so really no promise we merge peculiar bank format CNAB stuff at the OCA level. As for the upper payment abstractions, they'll move between 8.0 and 9.0 sadly.

rvalyi:
@mileo also, in case KMEE would be open to buy us such a dev, I could also get it done faster even for 8.0. Else we will end up doing it but no time-frame commitment right now sorry.

rvalyi:
@mileo I have seen your boleto's work and as far as I looked it doesn't look bad. You would help easy compatibilty and avoid mappings using the exact same field names of brcobranca in case you have choice. If you don't have such choice that's fine we will work around it.
cc @renatonlima

rvalyi:
:-1: So in fact, aside from needing tests, this module will need to depend on l10n_br_account_service and generate a service invoice I believe. We will double check this but I feel some more work is required.

rvalyi:
superseeded by https://github.com/OCA/l10n-brazil/pull/246
Closing; please shout if this is wrong.

PR #176, opened 2015-09-07
Adding gitter badge

danimaribeiro:


rvalyi:
I would say let's merge at least account_product before that otherwise we will defocus too much in having to explain the state of the project many times to users. But after that yes.

rvalyi:
@danimaribeiro I was not really in favor, but in fact the OCA is pretty opposed to the idea too: see the thread about this here:
https://odoo-community.org/groups/contributors-15/contributors-27060?mode=thread&date_begin=&date_end=

So at the moment I'm closing this. Once the project will have many modules merges in v9 I'll not be personaly hostile to the idea, but today I really feel it's loss of focus for now real benefit...

PR #185, opened 2015-10-10
Fix modules version number

danimaribeiro:
Fix #183

PR #187, opened 2015-10-10
better description of the project and its wider context

rvalyi:
Entao eu expliquei melhor o contexto do projeto, apontando mais projetos de extencao do team odoo-brazil.
Eu tambem botei o link da Truscode do @danimaribeiro que antes nao tinha site ativo.

Ok, com isso?

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,13 +76,13 @@ Available addons
 ----------------
 addon | version | summary
 --- | --- | ---
-[l10n_br_account](l10n_br_account/) | 8.0.0 | Brazilian Localization Account
-[l10n_br_base](l10n_br_base/) | 8.0.0 | Brazilian Localization Base
-[l10n_br_crm](l10n_br_crm/) | 8.0.0 | Brazilian Localization CRM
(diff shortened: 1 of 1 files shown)


danimaribeiro:
Ok 👍

rvalyi:
OK, so I added a diagram of the several related projects and their dependencies to help people understand the big picture. I also fixed the module numbers in the readme as discussed in https://github.com/OCA/l10n-brazil/pull/185

Exceptionally I'm proceeding with this trivial commit without waiting more approvals to avoid a mess with the 9.0 branch tomorrow.

danimaribeiro:
There are few typos, I can fix that later.

The diagram is really good.

rvalyi:
@danimaribeiro if you could help with the typos now that would be cool. Tomorrow they will create the 9.0 branch but before that they will also git mv the **uported** modules to the root (the **uported** terrorist system has been dropped for the 9 migration). As much as possible I would like to avoid spagethi rebases between 8.0, 9.0 and the current PR's because I fear it may end up generating conflicts (but may be that's just paranoia).

PR #188, opened 2015-10-12
Fix typos

danimaribeiro:


rvalyi:
:+1: thanks @danimaribeiro 
Quickly merging that before the mass commit storm.

PR #206, opened 2015-11-02
[WIP] Edoc refactory

mileo:
Work in progress, do not merge:
- [x] Generic workflow for eletronic governament invoice and normal invoice;
- [ ] Support manual export of l10n_br_account_product txt / xml invoices;
- [ ] e-doc example for abrasf nfs-e
- [ ] Support direct transmission of nfe https://github.com/odoo-brazil/odoo-brazil-eletronic-documents


mileo:
@renatonlima Esse é o erro q citei no gitter esses dias: https://travis-ci.org/OCA/l10n-brazil/jobs/88712660#L195

rvalyi:
@mileo,

There are several things involved here. Some changes will be easier to migrate in production than others. So basically we think the highest priority is to move the workflow customization from l10n_br_account_product back to l10n_br_account to have it generic for the NF-SE too.

The earlier we have that one, the less databases we will need to migrate...

As for the other changes it may take more time. And we are not even sure how much we will be able to achieve in v8 or v9.

Basically we would also like to make it work with the OCA connector for asynchrone transmission (NFe and NFSE RPS). I participated to the magento connector, built the solr (solerp) and devise connectors using connector so I am comfortable for doing this quickly. The problem is more that if we depend on connector, then we may delay the v9 migration quite a bit because connector may not be ported to v9 this soon https://www.indiegogo.com/projects/odoo-connector-odoo-magento-connector-for-odoo-9

So you so there are still some open questions.

Now if you could contribute moving the workflow customization as a 1st step that would be appreciated.

cc @renatonlima

danimaribeiro:
What are the pros in using  the connector to implement async?
It's a big dependency for something simple!

rvalyi:
pros:
- super reliable
- would make a reliable log system with replay ability
- this better historization and replayability would lower support costs (we really saw that when the Magento connector started using it and today these failed electronic invoice transmission directly translate in support costs).
- would provide nice abstractions to plug different transmission systems
- these core abstractions would make it viable to use a client-server SOAP transmissor, possibly reusing more mature and more maintained libraries with no need for these libs to be Python and in the same version as Odoo (SOAP is typically better implemented in Java or C# the languages it was designed for). I probably discovered a memory leak with the pysped transmissor, that would free us of that kind of troubles...

cons:
- big dependency
- might delay v9 migration if we depend on it

All in all, we should probably go without connector in v8. But if we really want to create a nice electronic invoice abstraction, then in v9 we may bite the bullet. It's probably possible to design it in a way were the connector is pluggable as an optional module. Not really a top priority but good to keep in mind. Just to give you an idea, some people (Acsone I believe) start depending on connector to export and import bank files in batch in a reliable fashion...

rvalyi:
BTW, I'm still no comfortable with v9. Many regression are not even considered. OCA big players are not using it in any critical project and even Odoo SA is not using its own dogfood. That one can be easily verified with:

```
curl --data-binary '{}' -H 'Content-Type: application/json;' https://www.odoo.com/web/webclient/version_info
```

Sure the design will end up improving things, but I'm just really not sure when it will be production proof if nobody is doing serious project with it...

PR #211, opened 2015-11-15
[NEW] README.rst

mileo:
Add readme to modules that are in 8.0 version

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,20 +76,20 @@ Available addons
 ----------------
 addon | version | summary
 --- | --- | ---
-[l10n_br_account](l10n_br_account/) | 8.0.1.0.0 | Brazilian Localization Account
-[l10n_br_base](l10n_br_base/) | 8.0.1.0.0 | Brazilian Localization Base
-[l10n_br_crm](l10n_br_crm/) | 8.0.1.0.0 | Brazilian Localization CRM
-[l10n_br_crm_zip](l10n_br_crm_zip/) | 8.0.1.0.0 | Brazilian Localization CRM Zip
-[l10n_br_data_account](l10n_br_data_account/) | 8.0.1.0.0 | Brazilian Localisation Data Extension for Account
-[l10n_br_data_base](l10n_br_data_base/) | 8.0.1.0.0 | Brazilian Localisation Data Extension for Base
-[l10n_br_zip](l10n_br_zip/) | 8.0.1.0.0 | Brazilian Localisation ZIP Codes
+[l10n_br_account](l10n_br_account/) | 8.0.2.0.0 | Brazilian Localization Account
+[l10n_br_account_product](l10n_br_account_product/) | 8.0.3.0.0 | Brazilian Localization Account Product
+[l10n_br_base](l10n_br_base/) | 8.0.2.0.0 | Brazilian Localization Base
+[l10n_br_crm](l10n_br_crm/) | 8.0.2.0.0 | Brazilian Localization CRM
+[l10n_br_crm_zip](l10n_br_crm_zip/) | 8.0.2.0.0 | Brazilian Localization CRM Zip
+[l10n_br_data_account](l10n_br_data_account/) | 8.0.2.0.0 | Brazilian Localisation Data Extension for Account
+[l10n_br_data_base](l10n_br_data_base/) | 8.0.2.0.0 | Brazilian Localisation Data Extension for Base
+[l10n_br_zip](l10n_br_zip/) | 8.0.2.0.0 | Brazilian Localisation ZIP Codes
 
 Unported addons
 ---------------
 addon | version | summary
 --- | --- | ---
 [l10n_br_account_payment](l10n_br_account_payment/) | 7.0 (unported) | Brazilian Localization Account Payment
-[l10n_br_account_product](l10n_br_account_product/) | 7.0 (unported) | Brazilian Localization Account Product
 [l10n_br_account_product_service](l10n_br_account_product_service/) | 7.0 (unported) | Brazilian Localization Account Product and Service
 [l10n_br_account_service](l10n_br_account_service/) | 7.0 (unported) | Brazilian Localization Account Service
 [l10n_br_account_voucher](l10n_br_account_voucher/) | 7.0 (unported) | Brazilian Localization Account Voucher
--- a/README.md
+++ b/README.md
@@ -76,14 +76,14 @@ Available addons
 ----------------
 addon | version | summary
 --- | --- | ---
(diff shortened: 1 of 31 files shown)


danimaribeiro:
I see that some repos the contributor list is done automatically.
Is it possible to do it here too?

rvalyi:
Hello @danimaribeiro
I prefer we do it manually, so we put the bar a bit higher. I don't think it's fair to include a contributor who did say 1% of the code or of the commits at the same level of visibility of somebody who did 20% or more. Think that if we do that automatically, in 1 or 2 years you'll have companies in Brazil that will make us loose our time with just one line contribs to get their name in the list (unfortunately we see this kind of behavior already here). So I prefer that the meritocracy is administered by the project leaders as it's the case in many projects. If somebody wants to show he did 2 or 3 commits in the repo, he can always show that here https://github.com/OCA/l10n-brazil/graphs/contributors which gives a fair picture of the contributions (it's using the main project branch though, meaning it will include last commits when they are forward ported to 9.0)

Also, as for yourself, if you contrib the merge of your work in l10n_br_account_service to be able to use it as a base for the NF-SE we will certainly put you as a contributor and co-author of the module. Also, reciprocally you can see that Akretion easily have 10 guys contributing to the OCA weekly and we don't ask as being listed as contributors at each commit we do either. Does it sound fair?

Regards.

rvalyi:
@danimaribeiro also we will propose to update the list of contributors in the modules of this PR. In any case, giving visibility to the main contributors of the project is absolutely our policy to encourage real contributions, that's also what we do clearly in the project README.

danimaribeiro:
I think every contribution is worth it.
But it's a matter of opinion.

" I don't think it's fair to include a contributor who did say 1% of the code or of the commits at the same level of visibility of somebody who did 20% or more"  
Maybe that contributor fixed a bug that no one has ever fixed for a year or more.

And also putting the name of each one that contribute is a matter of show that their little contribution is also important to the project and probably motivates them to contribute more.

rvalyi:
@danimaribeiro you know like us that there are already companies that do nothing in the project but try to control its image right? (I won't give names and start flame wars here). And that's not the only time in these 6 years either this happens... I'm 100% sure if one 1 commit = added in contributors, we will have some of these companies putting 5 guys doing 1 line commit each just to appear as being leading the stuff. Not fair. 
And we don't want "pagar para ver". If we didn't observe this opportunist behavior, probably we wouldn't put that entry barrier. For something less market oriented, like some infrastructure modules, they have little this problem, but in the Brazilian project this is a bit tough sadly...

Like in every open source project Project leaders are free to eval the usefulness of a contrib and the effort. We also know perfectly that we should give some visibility to the contributors it's exactly for this reason we list you or Luis as the main contributors of the project in the README. Again, smaller contributions are totally visible from the Github contributor page and we even point to that page in the README file so it's really not like if their participation is hidden. Not at all.

Regards.

rvalyi:
@danimaribeiro meanwhile for the guys who really want to help the project there are so many things yet to do like: stabilize HR modules, POS modules, all the NF-SE electronic invoices, all the payment gateways, the SPED...
So that's really not like if there is no space left for serious contributors to show their skills... There is no need to try highjacking other's projects for that...

BTW, we even think about promoting you and Luis as OCA PSC of these sub-projects based on your future contribs and if you are interested of course; cause we know we cannot eat it alone and it's not our interest either.

rvalyi:
@danimaribeiro other "bankable" OCA projects do no different and makes a difference between casual help and project pillars.  See for instance the Magento connector http://odoo-magento-connector.com

rvalyi:
@danimaribeiro also when you say "I think every contribution is worth it." It's really not that we think different.

It's rather that we don't want to make it trivial to game the project. The day you become PSC of the payment gateways and do 80% the gateways you probably won't like some company come, makes 3 commits and claim to author the project just like you and get the same visibility. Like the other day, Magno (Akretion) contributed to the commission project, we didn't try to convince the commission PSC to mention us a contributors just like the man authors https://github.com/OCA/commission/commit/ab15c692118daba890a3b2a9f10b14648b30ceac And this is true with the dozens of Akretion contributors and the dozens of OCA projects they participe weekly. The day we make 3 patches in some of your modules, we won't try to claim the same visibility as you either...

Basically we need to ensure the visibility goes to the companies who invest massively into the project. We don't sell a pro version of OCA/l10-brazil, our revenues comes directly from the prestige of doing it (consultancy), just like probably you will have your own visibility for the parts you will contribute massively (payment gateways and NF-SE may be?). In a context where Odoo SA is ready to sell visibility to any company giving them a check and claim a company like "Webkul is the best community contributor" (From Fabien today), I think that's important to bring back some meritocracy. It's certainly a matter of making this sustainable rather than not valuing the causal contribs.

mileo:
me neither https://github.com/OCA/commission/commits?author=mileo

But i think that you two have to review every module and see who contribute. Creating some view ou import csv/xml data is very easy. Make a improve in the tax engine, improve workflows, refunds ....
are countless hours.

This is the first time that i do this: https://github.com/OCA/l10n-brazil/pull/240#discussion_r45800008
You never see me in any place saying that this and that in any project. And if you see in my contrbutionsl10n-brazil it is very rare that I put my name in the header.

I never do like this @renatonlima https://github.com/OCA/l10n-brazil/commits/9.0/l10n_br_account_product/wizard/l10n_br_account_document_status_sefaz.py
Do some renames and put his name.

**Be altruistic, because comunity is bigger than code lines!**

rvalyi:
@mileo we never complained about your behavior either. I was only answering to @danimaribeiro about his specific suggestion to automatically count as contributor anybody doing a commit. I think the case you mentioned with @renatonlima is an accident and we can fix it. You can see that we even carefully propagated even Proge copyrights, while they didn't exactly had this behavior when they tried to fork the project before exploding some years ago... We also often mention you as a module contributor that is we do account for these numerous hours you are talking about. This doesn't change either that a guy like @renatonlima did even order of magnitude more work in the project since many years before.

We have no problem at all being altruistic with the community. I don't know if you see it but we are kind of working like mad to they get these PR's merged before v9 migration. This is the only reason we don't spend hours these days helping new comers in the list/chat as we were doing the years before (like when we offered you free training).  We also value your reviews and this is the only reason these PR's are open since months now instead of being merged.

BTW, I was suggesting future project responsablities for @danimaribeiro. As for you we are thinking about leading the HR and POS modules in the OCA, again if you agree. Today Kmee has a clear leadership there, while Truscode as some advance in the NFSe and payment gateways. If there is quality in these modules and project management, we can bring them to the OCA either in the main repo, either in other repos (depending on their critical mass and dependencies).

regards.

PR #213, opened 2015-11-15
[FIX] CamelCase Classes

mileo:


rvalyi:
@mileo looks good but then please update the module number to 8.0.3.0.0 as it could have some impact on sub-classes. Also would be cool if you could take advantage of this change to update the project README.rst to put l10n_br_account_product in the list of migrated modules. Thanks.

mileo:
@rvalyi Done, but modules in version 8.0.1.0.0 i change to 8.0.2.0.0

rvalyi:
@mileo thanks but please amend your last commit as we now have 2 flake8 errors: https://travis-ci.org/OCA/l10n-brazil/jobs/91182670#L755
Also, please include l10n_br_account_product and update the list of versions in this README section:https://github.com/OCA/l10n-brazil/blob/8.0/README.md#available-addons

Also, in case you could, in fact I did an error, that the 5th number of the module version that should have been incremented, not the 3rd as it doesn't require data migration nor module update. That's not a big deal we will merge it even if you don't change it. But if you could... Sorry for the wrong information before. See https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

mileo:
PEP8 FIXED in another PR #220

PR #221, opened 2015-11-15
8.0 oca merge mileo test pass

mileo:
Group PR to make test pass:
#220
#219
#218
#217
#216
#215
#214
#213
#212
#211

PR #223, opened 2015-11-15
8.0 oca freight ait

mileo:
@renatonlima 
In PR #222 the aproach that works actualy in https://github.com/odoo-brazil/l10n-brazil/

The way that we are discussing in gitter last night is here, but don't works. If you imagine a different approach let me know


mileo:
Depends of #221

mileo:
@renatonlima I've implemented another way and it worked

Some like that:  

```
@api.multi
def button_reset_taxes(self):
    result = super(AccountInvoice, self).button_reset_taxes()

    account_invoice_tax = self.env['account.invoice.tax']
    ctx = dict(self._context)
    for invoice in self:
        account_invoice_tax.create(
             {
                      ...
             })
    return result
```

PR #231, opened 2015-11-16
[FIX] ODOO-BRAZIL -> OCA: Corrige o workflow quebrado da SO e PO

mileo:
https://github.com/odoo-brazil/l10n-brazil/pull/56


renatonlima:
:+1:

rvalyi:
:+1: makes sense indeed.

PR #146, opened 2015-06-09
Migration/l10n br sale product

renatonlima:

PR #236, opened 2015-11-20
[FIX] Wrong method compute_all parameters

mileo:


rvalyi:
@mileo good catch
we got it right here https://github.com/OCA/l10n-brazil/pull/236/files#diff-9549f5a1703732ec8c0a1bd2ec94f6f3L251
and wrong here https://github.com/OCA/l10n-brazil/pull/236/files#diff-9549f5a1703732ec8c0a1bd2ec94f6f3L69

so :+1: 
cc @renatonlima

mileo:
But I still can not save

rvalyi:
@mileo are you testing the module alone?
I'm not sure the module alone is useful or if you could have it alone. Because basically you'll always have l10n_br_sale_product or l10n_br_sale_service installed via auto_install.

We tested with these modules before submitting the l10n_br_sale module so we never tried standalone as it's a configuration that would not exist once the next PR's are processed. May the issue you are facing comes from this?

mileo:
I'm testing alone, but is working in a new database now. Sorry for the wrong error post. 

But this error make me debug and found this one.

rvalyi:
@mileo I'm rebasing https://github.com/OCA/l10n-brazil/pull/146 on 8.0 right now for review, that may also help you for testing.

mileo:
=( i'm having the error again.

But i can't see why.

Occurs after i try to create a fiscal category with a user that don't have the right permissions

renatonlima:
:+1:

PR #238, opened 2015-11-20
[ADD] CRM tests.

mbcosta:


rvalyi:
:+1: LGTM, thanks @mbcosta for testing that code that was untested so far.

renatonlima:
:+1: @mbcosta You can do other PR to add test to check CNPJ and IE validation

PR #242, opened 2015-11-24
comment some useless tests until migration is finish

renatonlima:

PR #243, opened 2015-11-24
comment some useless tests until migration is finished

rvalyi:
accorde Travis...


rvalyi:
era para acontecer,
o Travis travou...

rvalyi:
ok build started in https://github.com/OCA/l10n-brazil/pull/242
seems that we are overflowing Travis a bit today...

PR #245, opened 2015-11-27
Zip search with correios' webservice

mileo:
filter-branch of https://github.com/odoo-brazil/odoo-brazil-addons


mileo:
https://coveralls.io/github/odoo-brazil/odoo-brazil-addons?branch=8.0
86% de coverage

danimaribeiro:
Do you want to implement the search by street address?
You can copy from here if you want:
https://github.com/Trust-Code/trust-addons/blob/8.0/trust_search_cep/models/l10n_br_zip.py#L57

mileo:
@danimaribeiro I think that correios webservice don't have this type of request
cc @mstuttgart

mstuttgart:
@danimaribeiro correios webservice don't have this service. The search is done only by zip code.

mileo:
@rvalyi please tag as new feature

danimaribeiro:
Check the code in the link.

mileo:
@danimaribeiro I see but you are using "http://viacep.com.br/ws". Our intent is only use official and free services. Unfortunately the correios WS do not have the search name option.

danimaribeiro:
Cep search is not that important, so doesn't need to be official.

rvalyi:
@mileo @danimaribeiro I agree we better use the official an free services in the OCA at least for now and leave these extensions to extra modules.

In any case we will never have consensus on all modules and potential extra services to use. So we better focus on a smaller scope where this is doable (hey but we are still the most complex OCA localization). Nearly all OCA localization work this way BTW.

danimaribeiro:
Three lines of code guys:

```
        url_viacep = 'http://viacep.com.br/ws/' + city.state_id.code + \
            '/' + city.name + '/' + street + '/json/unicode/'
        obj_viacep = requests.get(url_viacep)
        results = obj_viacep.json()
```

mileo:
For me is a option that we don't need to have in Odoo. 

If zip do not return a address user should use the Correios website.

Keep many of zip wizards have a hight cost and a low return.

rvalyi:
@danimaribeiro well ok for 3 lines of code I change my mind then. We should be fault tolerant of the service is down but yeah, we can include these 3 lines of code.

mileo:
:+1: Be fault tolerant of the service.
:-1: To search by address.

rvalyi:
@mileo,

we can have an ir.property allowing or disabling the viacep check then. That's still very few lines of code for the feature. I also approve if the service is fault tolerant: like having a timeout and not exploding if viacep timeouts.

danimaribeiro:
Guys, please look the code in the link.

mstuttgart:
I believe that is better keep _l10n_br_zip_correios_ and _l10n_br_crm_zip_correios_ doing search zip code from correios webservice and add this _street address search_ in other module.

rvalyi:
@danimar, do you believe you could add street support in around 5 lines or does it change the request and response API?
If it's really so simple, I think we can consider having it as an option, if not as an extra module. Now, may be you have hooks to suggest so that to help plug the viacep search in case that should be an extra module.

my 2 cts.

mileo:
https://github.com/jbochi/cep/blob/master/cep/__init__.py

l10n_br_crm_zip_correios/README.md
```diff
@@ -0,0 +1,3 @@
+Módulo Para Busca do CEP no CRM através do webservice dos correios
```
renatonlima, line 1: Please remote this file.

danimaribeiro:
An approach like this would be better:
https://github.com/odoo-brazil/odoo-brazil-addons/pull/15

Doesn't need to add a new module every time we need a new ZIP field.

renatonlima:
@danimaribeiro 

Eu vejo 2 opções:
1. Manter módulos para implementar consultas diferentes (l10n_br_zip_correios, l10n_br_zip_viacep e etc..)
2. Ter apenas um módulo como l10n_br_zip_search e implementar a factory para escolher o provedor que deseja utilizar para consultar os ceps (viazip ou correios) desta forma para cada novo provedor é só implementar a classe dele no módulo.

O que acha?

@mileo @rvalyi

danimaribeiro:
A ideia era ficar mais simples pra diminuir a quantidade de módulos. Implementar uma factory?
A questão principal que quis levantar não é em relação a qual webservice utilizar e sim, a sobrescrever o método mais apropriado para evitar que quando tenhamos um novo campo zip lá em um módulo de estoque, tenhamos que criar um l10n_br_zip_stock por exemplo.

renatonlima:
@danimaribeiro, 

Quando falei em implementar uma factory, foi para resolver a questão dos provedores de pesquisa de CEPs, sobre a sua abordagem eu não entendi muito bem, porque os campos de endereços usados no CEP fica no base, l10n_br_base, só foi criado o l10n_br_zip_crm para adicionar o botão de pesquisa no crm.lead, caso exista outro módulo onde você queira implementar este botão deveria fazer a mesma coisa. Você pode explicar melhor a sua abordagem?

PR #251, opened 2015-12-07
Migration odoo-brazil to OCA of l10n br base

mileo:


rvalyi:
the migration script itself will do no harm, but openupgrade_analysis.txt is a bit problematic as normally we would expect a file related to 7.0 or eventually the develop branch here, not some other arbitrary 8.0 branch. Eventually this may clash if later somebody does scripts to migrate from v7 (we might do it at least for some core modules). So eventually I propose to commit only the migration file or alternatively keep that work in some other branch.

mileo:
As Pedro Baeza said here: https://github.com/OCA/OpenUpgrade/issues/416#issuecomment-162852204

We can add openupgrade as a dependency. 

About 7.0 i'm testing it, but i think that will work too.

rvalyi:
@mileo I also think it will work too with 7.0 that why I said "it will do no harm". My problem is with the openupgrade_analysis.txt files which are very specific to the arbitrary branch you are migrating from while normally the analysis against the previous serie is expected here (7.0 or most probably the development one in our case).

I would say:

A) Either you leave these specific migration files in some migration technical branch in the odoo-brazil repo as it's more related to that repo than the OCA repo.

B) Either we accept the migration files, but then only the .py files. not the openupgrade_analysis.txt files. And with a comment making it explicit that this is to migrate from that branch specifically.

If everybody start submitting their migration files from their temporary work branches the project would quickly become a mess. I'm open to accept exceptionally for this one as several people used that temp branch, but then at the condition it will not conflict with the migration from the 7 serie. As for adding the openupgrade dependency I don't think we should do it, the other repos don't do it either I believe.

rvalyi:
one more comment: in the past our project had twice as less contributors and openupgrade wasn't as easy. That is why for instance we created the "development" branch after the 7.0 branch and before the 8.0 branch. So we didn't break the 7.0 backward compatibility.

In the future, with a careful module versioning, we will be able to do non backward compatible changes and provide a migration script between versions. This is how I propose to deal with moving the invoice workflow extension in l10n_br_account for instance.

PR #268, opened 2015-12-31
8Oca/fix/tags exportacao

mileo:


renatonlima:
this PR is wrong, the right PR is #269

PR #277, opened 2016-01-29
[FIX] Wrong function parameters

mileo:


renatonlima:
@mileo please rebase and fix pep8

mileo:
@renatonlima rebased and pep8 cleared

rvalyi:
@mileo, sorry if I'm wrong about it, but I fail to see where is the bug before your patch. From the dependencies I see onchange_product_id defined here
https://github.com/odoo/odoo/blob/8.0/addons/stock/stock.py#L2129
and called with the proper arguments in the existing code.

Is there a special use case where this code would fail? Another module combo may be?

Now I then to agree it's more robust (change proof) to call the super onchange_product_id with keyword arguments. But if we do so, then probably it's also cleaner to have our own onchange_product_id method with keyword arguments as well here https://github.com/OCA/l10n-brazil/pull/277/files#diff-c1261c5d7b84bf43e71d5348e2cdf200L123

What do you think? Could you please give more details about the current issue? Thanks.

danimaribeiro:
I agree that parameters seems to be right before the PR.

But :+1:  to use as the original method and merge this, it does not harm.

rvalyi:
@mileo I vote on this if you also set keywords arguments here https://github.com/OCA/l10n-brazil/pull/277/files#diff-c1261c5d7b84bf43e71d5348e2cdf200L123
as I said in my previous comment.

mileo:
Pelo que eu me lembro havia um erro neste trecho quando outro modulo sobrescrevia o método.

Tentei simular aqui mas faz tanto tempo que eu não me lembro, desculpe.

PR #279, opened 2016-02-12
The field price_total was duplicated in model account_invoice_line.

mbcosta:
O campo price_total estava duplicado no objeto account_invoice_line


renatonlima:
:+1:

PR #291, opened 2016-03-29
[MIG] Change the version of module.

mbcosta:


danimaribeiro:
:+1: 
Just the building is failling, but because the travis points to branch 9.0 of Pysped which does not exist. Can you change to 8.0 in this PR to make it pass

danimaribeiro:
I made a patch:
https://github.com/OCA/l10n-brazil/pull/300
Let's see if it fix the build, then we can start from there.

renatonlima:
:+1:

mileo:
:+1:

PR #299, opened 2016-04-26
Add xPed and nItemPed fields to NFe

leorochael:
Add optional `xPed` field to NFe invoice header with with information
from `accounting.invoice` field `name`.

Add optional `xped` and  `nitemped` fields to invoice lines and copy the
info to `xPed` and `nItemPed` product fields in the NFe.

### DIFF
--- a/l10n_br_account_product/models/account_invoice.py
+++ b/l10n_br_account_product/models/account_invoice.py
@@ -23,6 +23,7 @@ from lxml import etree
 from openerp import models, fields, api, _
 from openerp.addons import decimal_precision as dp
 from openerp.exceptions import RedirectWarning
+from openerp.exceptions import ValidationError
 
 from openerp.addons.l10n_br_account.models.account_invoice import (
     OPERATION_TYPE,
@@ -756,6 +757,21 @@ class AccountInvoiceLine(models.Model):
     freight_value = fields.Float(
         'Frete', digits=dp.get_precision('Account'), default=0.00)
     fiscal_comment = fields.Text(u'Observação Fiscal')
+    xped = fields.Char(
+        string=u"Código do Pedido (xPed)",
+        size=15,
+    )
+    nitemped = fields.Char(
+        string=u"Item do Pedido (nItemPed)",
+        size=6,
+    )
+
+    @api.onchange("nitemped")
+    def _check_nitemped(self):
+        if self.nitemped and not self.nitemped.isdigit():
+            raise ValidationError(
+                _(u"nItemPed must be a number with up to six digits")
+            )
 
     def _amount_tax_icms(self, tax=None):
         result = {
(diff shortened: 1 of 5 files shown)


l10n_br_account_product/models/account_invoice.py
```diff
@@ -756,6 +757,21 @@ def _compute_price(self):
...
     fiscal_comment = fields.Text(u'Observação Fiscal')
+    xped = fields.Char(
+        string=u"Código do Pedido (xPed)",
+        size=15,
+    )
+    nitemped = fields.Char(
+        string=u"Item do Pedido (nItemPed)",
+    )
+
+    @api.onchange("nitemped")
```
danimaribeiro, line 769: onchange or constraint

leorochael, line 769: @danimaribeiro, `onchange` because I don't want the user to have to wait until trying to click "save" to see that the `nItemPed` field is in the wrong format.

danimaribeiro, line 769: Why not use Integer field for that? And why just this field is important to have a specific onchange, the user can type wrong values in any field.

leorochael, line 769: I tried using `field.Integer`, but it defaults `0` instead of being empty.

The `nItemPed` tag is explicitly optional in the NFe definition. Having the `nitemped` field default to `0` would have the consequence of there always being an `nItemPed' tag in the XML.

There is nothing special about `nitemped`, but it would be unfriendly to wait until the user tries to save the object to notify about the wrong value of what is essentially a very optional field, hidden behind an "Extra Information" tab, forcing the user to hunt through all tabs until it finds the offending field.

If you believe other fields could also benefit from more immediate feedback about their wrong values, then I encourage you to add `onchange` validators for them as well.

danimaribeiro, line 769: Better to use "Integer" then and fix the tag in the pysped to "Inteiro"
https://github.com/odoo-brazil/PySPED/blob/8.0/pysped/nfe/leiaute/nfe_200.py#L1443

leorochael, line 769: I agree the tag in PySPED might be needed to be changed to `TagInteiro`, in which case we would need to change the assignment code in `document.py` to make sure it's sending an integer if the value is present in the invoice line.

However this is completely unrelated to why I chose to use `field.Char` instead of `field.Integer`: the field is optional, and should be empty when presented to the user, instead of containing a `0`.

If you want, I can add a comment to the code to make this clear.

l10n_br_account_product/models/account_invoice.py
```diff
@@ -756,6 +757,21 @@ def _compute_price(self):
     freight_value = fields.Float(
         'Frete', digits=dp.get_precision('Account'), default=0.00)
     fiscal_comment = fields.Text(u'Observação Fiscal')
+    xped = fields.Char(
```
renatonlima, line 760: Eu renomearia este campo para algo do tipo customer_ref ou customer_order

PR #314, opened 2016-05-19
Internal_number in list view of invoice

sadamo:


renatonlima:
:+1:

rvalyi:
:-1:  replacing a field with another is not a good practise I think. The reason is that some module could expect this field to be present to compute an other field or a color, I remember we had views errors in the old days of the localization when we were doing that kind of things... I prefer an extension that would set this field as invisible and add the internal_number after or before it.

sadamo:
@rvalyi, good point. I've altered that and squashed the commit.

rvalyi:
:+1:  thanks @sadamo

mileo:
:+1:

renatonlima:
:+1: @rvalyi good catch!

l10n_br_account/models/account_invoice.py
```diff
@@ -43,6 +43,7 @@
 
 class AccountInvoice(models.Model):
     _inherit = 'account.invoice'
+    _order = "internal_number desc, id desc"
```
danimaribeiro: We should find a better way to sort this, as internal_number is a char field, this does not work as expected.
broken order

rvalyi:
@sadamo could you please rebase on 8.0 ? I also think @danimaribeiro made a good point...

l10n_br_account/models/account_invoice.py
renatonlima: I think we can order by date_intoice and internal_number

sadamo: Maybe date_hour_invoice and internal_number?

renatonlima: @sadamo why you add _order property in l10n_br_account and l10n_br_account_product, I can add _order property only in l10n_br_account is enough.

sadamo: @renatonlima because the field date_hour_invoice is declared in l10n_br_account_product. But if using date_invoice to order, which now I think should be the best, the l10n_br_account_product  _order can be removed.

mileo:
Rebased

l10n_br_account/models/account_invoice.py
renatonlima: Best way is to use date_invoice , because in account.invoice CAN be used to create other fiscal document types like NFS-e, using date_invoice, you can remove _order in l10n_br_account_product

mileo:
:+1: again

l10n_br_account/models/account_invoice.py
sadamo: Something strange happened when I use the _order property only in l10n_br_account the way we are discussing, it doesn't update the view, even in new db. 
Only in l10n_br_account_product this work. Anyone knows something about?

l10n_br_account_product/models/account_invoice.py
```diff
@@ -37,6 +37,7 @@
 
 class AccountInvoice(models.Model):
     _inherit = 'account.invoice'
+    _order = "internal_number desc, date_hour_invoice desc"
```
renatonlima: To correct order you must invert the order of columns: _order = "date_hour_invoice desc, internal_number desc"

l10n_br_account/models/account_invoice.py
renatonlima: @sadamo the problem is in the line 127 there is another definition of the attribute _order https://github.com/kmee/l10n-brazil/blob/916ad3cf2e80d1e755972034ddc8943f4d34c061/l10n_br_account/models/account_invoice.py#L127

You can remove the line 127 and change _order atribute to: _order = "date_invoice desc, internal_number desc" and in l10n_br_account_product chante to _order = "date_hour_invoice desc, internal_number desc"

sadamo: I only used the _order property in the l10n_br_account with: "date_invoice desc, internal_number desc". Because if used the date_hour_invoice there will not be an sort, I think it will became confuse.

sadamo:
Altered the _order to "date_invoice desc, internal_number desc". Ready to merge?

rvalyi:
@renatonlima @sadamo didn't we agree to override the _order and use date_hour_invoice in l10n_br_product for more precision and minimize possible order errors in case of several invoices the same day with the string sorting issue raised by @danimaribeiro ?

sadamo:
I have made some tests and with date_hour_invoice in _order the list will have no sorted sequence, I think the date_invoice is better. In the firsts numbers we can use the ir_sequence to complete with zeros. What do you think? @renatonlima @danimaribeiro @rvalyi

danimaribeiro:
I don't get it:
"date_hour_invoice in _order the list will have no sorted sequence"

renatonlima:
Hi guys,

@sadamo as I said in l10n_br_account_product we can keep _order = 'date_hour_invoice DESC, internal_number DESC' to minimize possible order errors in case of several invoices the same. I think it is enough.

sadamo:
Ops, my bad, I mixed with another feature. Will change to date_hour_invoice.

sadamo:
Done!

renatonlima:
LGTM :+1:

rvalyi:
:+1: thanks @sadamo!

mbcosta:
:+1:

PR #334, opened 2016-06-17
search invoices by its internal_number without wildcards

fernandomr:
Adds invoices search by their exact number instead of searching by parts of it using odoo wildcards.


rvalyi:
:+1:

mileo:
:+1:

renatonlima:
:+1:

PR #335, opened 2016-06-21
Feature/account voucher baixa bonificacao

fernandomr:
Moves wizard from l10n_br_account_voucher to l10n_br_account.
Inherits invoice_validate and adds behavior to create a voucher to invoices that don't have commercial value and must not count as client's debits.
Adds two fields on account.journal to determine either that's a remittance journal or not and if does the invoices in that journal must be reconciled in the act of it's confirmation.
Adds yml tests

replaces https://github.com/OCA/l10n-brazil/pull/333


fernandomr:
Replaced by https://github.com/OCA/l10n-brazil/pull/337

PR #341, opened 2016-06-22
[FIX] Header stock.picking fiscal fields.

mileo:
If you create a sale order to customer Contribuinte PE, invoice after delivery:
- Insert on item one: Mouse;
- Insert on item two: Ipad;

If you create a sale order to customer Contribuinte PE,  invoice after delivery:
- Insert on item one: Ipad;
- Insert on item two: Mouse;

And if you compare header fiscal fields they will be different.

### DIFF
--- a/l10n_br_sale_stock/models/stock.py
+++ b/l10n_br_sale_stock/models/stock.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+# (C) 2016 Luis Felipe Mileo - KMEE
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from openerp import api, models
+
+
+class StockMove(models.Model):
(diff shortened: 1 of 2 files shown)


l10n_br_sale_stock/models/stock.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class StockMove(models.Model):
+    _inherit = 'stock.move'
+
+    @api.multi
+    def action_confirm(self):
+        """
+            Pass the department_id and the dispatch_note to the picking)
+        """
```
rvalyi, line 16: looks like the docstring doesn't match what the code does, better remove it then.

mileo:
@rvalyi fixed

mileo:
@lfdivino Poderia revisar este bug, ele é super importante

PR #350, opened 2016-07-13
[FIX] Missing to pass domain.

mbcosta:
We missing to pass the domain, this error make wrong results because the program need to receive this domain with the Date to filter the moves and values. @sadamo @mileo


renatonlima:
:+1:

mileo:
:+1:

PR #355, opened 2016-07-20
[8.0] Included function amount_to_text for Brazilian Portuguese

mbcosta:
The right place to put this function is openerp/tools/amount_to_text.py and I make pull request https://github.com/odoo/odoo/pull/11388 , but Odoo take long time to analise and merged, so until we waiting we can put this on our localization.

### DIFF
--- a/l10n_br_base/__init__.py
+++ b/l10n_br_base/__init__.py
@@ -19,3 +19,4 @@
 
 from . import models
 from . import tools
+from . import tests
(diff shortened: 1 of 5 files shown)


rvalyi:
agree on the principle as I suggested it, but please fix the PEP8 https://travis-ci.org/OCA/l10n-brazil/jobs/146230268#L899

rvalyi:
:+1: thanks @mbcosta !

renatonlima:
:+1:

rvalyi:
@mbcosta that would be great also if you update your PR in Odoo core here https://github.com/odoo/odoo/pull/11388 with this latest code (don't bother including the tests in the core PR), We never know, but at least if Odoo SA merges it one day at least it would be the correct version (as you recently added a few fixes).

mbcosta:
@rvalyi done, now the both PRs have the same code.

PR #365, opened 2016-08-06
FIX unicode warning updating module

cmsalmeida:


rvalyi:
:-1: we should merge this first https://github.com/OCA/l10n-brazil/pull/290
Also it's likely that rebasing that PR, we will forward port your 8.0 patch, so I'm closing this PR. Thanks.

PR #389, opened 2016-08-18
migration 7.0 (old develop) -> 8.0 - l10n_br_crm

mbcosta:
Just needed to rename column cnpj_cpf to cnpj


rvalyi:
:-1: If I'm correct the old cnpj_cpf field was used to store either the cnpj or the cpf in 7.0. So here with your migration the new cnpj column will contain both cpf and cnpj records. We can keep your column rename but we would need an extra migration method to move the cpf records out of that cnpj column to th new cpf column.

For this we will probably need to use the same criteria as: https://github.com/OCA/l10n-brazil/blob/7.0/l10n_br_crm/crm_lead.py#L49

Note; B2B companie can safely use your PR, only B2C companies would corrupt their database.

mileo:
:-1: Concordo com o @rvalyi

PR #390, opened 2016-08-19
PEP8 on base and CRM module

mbcosta:
Replace depreciated api.one decorator and remove excess space on license header.


rvalyi:
:+1:

renatonlima:
:+1:

rvalyi:
@api.constrains needs an @api.multi indeed. But then their logic should then be wrapped in loops instead of having that incompatible ensure_one.

mbcosta:
Thanks @rvalyi for review, I change all constrains in base and crm

PR #391, opened 2016-08-20
Included requirements.txt file

mbcosta:
Because module l10n_br_zip_correios depends of suds python package.


mileo:
:+1:

renatonlima:
:+1:

rvalyi:
Just a note to say I'm not sure this brings anything as today Odoo modules are not Python modules and tools such as the OCA tooling use the `__openerp__.py` meta-information to converge that kind of dependencies already. If somebody has any pointer if it's a good practice or not in the OCA...

rvalyi:
@pedrobaeza @sbidoul @dreispt Hello. You guys are the OCA packaging experts I believe so I'm asking you: do we need this requirements.txt file aside from specifying the "suds" pypi external dependency inside the `___openerpo___.py` package already ? Does it bring any advantage here?
See https://github.com/OCA/l10n-brazil/blob/8.0/l10n_br_zip_correios/__openerp__.py#L22

sbidoul:
`requirements.txt` is indeed redundant with the `external_dependencies` key of `__openerp__.py` 

In practice it is required by OCA travis scripts to install dependencies because these scripts don't know how to interpret external_dependencies.

With the help of `setuptools-odoo` which make true python packages and https://wheelhouse.odoo-community.org, this is about to change. I just need to find the time to finish and test https://github.com/OCA/maintainer-quality-tools/pull/343

rvalyi:
@sbidoul thanks you answered exactly my questions! OK, then I :+1: for now until eventually this changes with the new wheel developments. Kudos for that work BTW, we are all eager to make our way out of the current Odoo package management hell!

rvalyi:
Gahhh, o test com suds rodou com o repo odoo/odoo e nao com OCA/OCB... Ainda nao olhei porque...
https://travis-ci.org/OCA/l10n-brazil/builds/154874137

rvalyi:
Deve ser um inconsistencia do webservice dos correios, madei rodar de novo para saber...

mileo:
Seria isso que esta travando o runbot?

rvalyi:
@mileo certeza que nao e. Parece que e inconsistente mesmo: rodei o Travis de novo e ele foi, deve ser problema la nos correios.

O problema do Runbot e outro porque nunca roda mesmo. Tou esperando a galera da Europa chegar de ferias direito e vou investigar com quem sabe bem disso.

PR #395, opened 2016-08-23
migration 7.0 (old develop) -> 8.0 - l10n_br_data_base

mbcosta:


renatonlima:
@mbcosta 

It seems that the two files are equal.

Issue #416, opened 2016-10-06
Migration to version 10.0


pedrobaeza:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0

Marque seu nome e o modulo que vc estiver migrando/revisando no documento:

https://docs.google.com/spreadsheets/d/1KnWbtAMyVoK-5uLsv3vv9Z8y7g5H6O-Tpu7E8bszxhY/edit#gid=0

# Modules to migrate
- [x] l10n_br_account
- [x] l10n_br_account_payment_order
- [x] l10n_br_account_product
- [x] l10n_br_account_product_service
- [x] l10n_br_account_service
- [x] l10n_br_base
- [x] l10n_br_crm
- [x] l10n_br_crm_zip
- [x] l10n_br_purchase
- [x] l10n_br_sale
- [x] l10n_br_sale_product
- [x] l10n_br_sale_service
- [x] l10n_br_sale_stock
- [x] l10n_br_stock
- [x] l10n_br_stock_account
- [x] l10n_br_zip
- [x] l10n_br_zip_correios

rvalyi:
guys, as nothing has been merged in 9.0 yet. The plan is to re-create the 9.0 branch very soon (and rebase the PR's against it). Then we will also rebase that 10.0 branch. Basically as soon as we have these structuring PR's merged: 
-  https://github.com/OCA/l10n-brazil/pull/386
-  https://github.com/OCA/l10n-brazil/pull/363

Thus we will minimize the discrepancy between latest 8.0 devs and 9.0 and 10.0. The other PR's have few commits and can be cherry picked without problem later. So basically if you want to speed the 10.0 migration please help us to get these important PR's in 8.0 first to avoid double working on these important topics.

pedrobaeza:
OK, just tell me and I re-run the 8.0 to 9.0 migration script, and when you're ready, the 9.0 > 10.0 one.

rvalyi:
@pedrobaeza yep, hopefully it will be ok early next week we will let you know. We are merging and reviewing at a good pace these days, these are the final miles of the important 8.0 refactors while the 9.0 has been dropped by many of us (still a few 9.0 projects with the open PR's done).

renatonlima:
Hello guys,

We finished migrating the v10.0 modules, some modules have not been migrated. Now we are working on version 12.0 migration and the goal of the l10n-brazil project is until the next one to keep location synchronized with Odoo CE version 14.0.

PR #439, opened 2016-10-24
migrate l10n_br_base, l10n_br_data_base to V10

yk2kus:


rvalyi:
@yk2kus thanks but please rebase on top of the existing PR's for 9.0 and even you'll still need to rebase again when we will re-create the 9.0 and 10.0 branches. So eventually it's better you help finishing the 8.0 rush or work on the upstream repos while we finish this.

rvalyi:
(with little changes the PR's for these modules of the 9.0 version works on 10,0 already. Also the priority is to the guys who first proposed the PR's so you cannot ask the 9.0 author to redo their work if you didn't analyse or critizice or use their previous work that had pending merges, I hope you understand.). It's also important that the 8.0, 9.0 and 10.0 use the same git "continuity" as much as possible. For instance may be migrating form 8.0 to 9.0 will involve making an homologation step to 9.0 (possibly for some non localization modules), so continuity is important here...

rvalyi:
@yk2kus also the rebase and migration thing was explained here:
https://github.com/OCA/l10n-brazil/issues/416#issuecomment-252282084

I'm really aware the current migration process sucks. But it's mostly a packaging issue: you see now everybody suddenly wake up and want to get PR into 8.0 before the split so the 9.0 and 10.0 branch preparation is lagging and this is fundamentally stupid, this is because all modules are in the same repo. The more we will split the modules by chunk (ideally one repo per module), the more we will be able to migrate stuff independently smoothly...

yk2kus:
@rvalyi Thanks, I understood. The thing was I checked if someone has already migrated to 9.0 or 10 but unfortunately couldn't find may be because the PR was too old and in the repo it was still uninstallable=False. I did double check now and found the PR for 9.0. :+1:

rvalyi:
@yk2kus  by the way I totally understand it sucks you know... Look, I'm always weighting to split the huge OCA repos like this one in smaller chunks... Had we done this, there would have been a ready to use 9.0 and even 10 branches already and you wouldn't have lost time and mostly double work (because what de did for 9.0 works 95% here in v10). But we are still including very large accounting refactors in v8. So we if we created the 9 branch for real before, we would have troubles properly forward porting these huge changes and managing the guys who would like to migrate the more advanced accounting modules to 9 or 10 meanwhile. All in all the trad off was to postpone the merges in the 9.0 until these essential accounting refactor make their way in the 8.0 branch... I hope you understand and also understand why breaking the localization into smaller chunks is essential...

PR #456, opened 2016-11-01
[8.0] Removed deprecated function amount_to_text replaced by Num2Words

mbcosta:
I tried add Brazilian Portuguese in function amount_to_text https://github.com/odoo/odoo/pull/11388 but the function is deprecated https://github.com/odoo/odoo/pull/12542#issuecomment-237590275 and the best option seems a python library Num2Words https://github.com/savoirfairelinux/num2words. While it not included in Odoo requeriments I put in our localization.


mbcosta:
To import the lib I use TRY function, the EXCEPT is log message in Odoo the first ideia is use Warning but I got error from Runbot, by this reason I just use INFO log

rvalyi:
:+1:

renatonlima:
:+1:

lfdivino:
:+1:

PR #529, opened 2017-04-10
prototipo nova versão da localização brasileira v10

mileo:
Comportamento esperado depois do PR:
------------------------------------
- [X] Módulo financeiro; https://github.com/OCA/l10n-brazil/pull/520
- [X] Módulo fiscal reestruturado; https://github.com/OCA/l10n-brazil/pull/518
- [X] Módulo de transmissão de nfe e nfce;

Roadmap:
https://docs.google.com/document/d/1JNJgapClBnDvpT-GphJyLrhLP5azn_D8g9TXjbESdB4/edit?pli=1#heading=h.jfitt9pmfp7d

Substitui o PR:
https://github.com/OCA/l10n-brazil/pull/527

Runbot para testes:
http://runbot.kmee.com.br/runbot/repo/git-github-com-odoo-brazil-l10n-brazil-wip-git-3

PR #544, opened 2017-06-06
[10.0] Merge odoo brazil wip/copy views

rvalyi:
see here for the extraction script #529 (comment)

### DIFF
--- a/copy_views/models/inherited_ir_ui_view.py
+++ b/copy_views/models/inherited_ir_ui_view.py
@@ -47,7 +47,7 @@ class IrUiView(models.Model):
 
         return res
 
-    @api.model
+    @api.multi
     def write(self, values):
         copied_views = []
 
@@ -86,7 +86,7 @@ class IrUiView(models.Model):
 
         return res
 
-    @api.model
+    @api.multi
     def unlink(self):
         copied_views = []
 
--- a/copy_views/models/inherited_ir_ui_view.py
+++ b/copy_views/models/inherited_ir_ui_view.py
@@ -41,10 +41,9 @@ class IrUiView(models.Model):
         #
         if values.get('inherit_id'):
             cv = self.search([('copy_id', '=', res.inherit_id.id)])
-            if cv:
-                arch = res.read_combined(['arch'])['arch']
-                for view in cv:
-                    view.write({'arch': arch})
+            arch = res.read_combined(['arch'])['arch']
+            for view in cv:
+                view.write({'arch': arch})
 
         return res
 
--- a/copy_views/models/inherited_ir_ui_view.py
+++ b/copy_views/models/inherited_ir_ui_view.py
@@ -5,14 +5,8 @@
 # License AGPL-3 or later (http://www.gnu.org/licenses/agpl)
 #
 
-from lxml import etree
 
-from odoo import api, fields, models, _
-from odoo.exceptions import ValidationError
-from odoo.modules.module import get_resource_from_path
-from odoo.tools.parse_version import parse_version
-from odoo.tools.view_validation import valid_view
-from odoo.tools.translate import encode
+from odoo import api, fields, models
 
 
 class IrUiView(models.Model):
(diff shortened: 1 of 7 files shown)


copy_views/models/inherited_ir_ui_view.py
```diff
@@ -0,0 +1,140 @@
...
+    def create(self, values):
+        #
+        # Ensures that the arch of the copied view is realy copied from the
+        # original view
+        #
+        if values.get('copy_id'):
+            copied_view = self.browse(values['copy_id'])
+            values['arch'] = copied_view.read_combined(['arch'])['arch']
+
+        copied_views = []
```
rvalyi: useless variable

copy_views/models/inherited_ir_ui_view.py
```diff
@@ -0,0 +1,140 @@
...
+
+        res = super(IrUiView, self).create(values)
+
+        #
+        # Ensures inheritance gets replicated to copied views
+        #
+        if values.get('inherit_id'):
+            cv = self.search([('copy_id', '=', res.inherit_id.id)])
+            if cv:
```
rvalyi: no need for if cv as cv will alway be iterable, for view in cv: is enough

copy_views/models/inherited_ir_ui_view.py
```diff
@@ -0,0 +1,140 @@
...
+                for view in cv:
+                    view.write({'arch': arch})
+
+        return res
+
+    @api.constrains('arch_db')
+    def _check_xml(self):
+        for view in self:
+            #
+            # When the view is being copied, there's no need to validate it's
```
rvalyi: it's -> its

copy_views/models/inherited_ir_ui_view.py
```diff
@@ -0,0 +1,140 @@
...
+            cv = self.search([('copy_id', '=', res.inherit_id.id)])
+            if cv:
+                arch = res.read_combined(['arch'])['arch']
+                for view in cv:
+                    view.write({'arch': arch})
+
+        return res
+
+    @api.model
+    def write(self, values):
```
rvalyi: should be @api.multi and there are other bugs, seems untested function

rvalyi:
Hello @aricaldeira. At Akretion we liked the idea of this module. I made e few comments and I think the write method is still screwed currently (wrong signature/not tested).

Could you please make a PR with this module against the https://github.com/OCA/server-tools project?

At the moment they would probably block your module because it has no unit test. It would be cool if you could add some... in the OCA/l10n_br_brazil project we could be a bit more tolerant to accept your module before it is accepted into the server-tools project. But at least it would be cool if we could get some review from the big OCA guys before using it here. Not to say it could benefit a lot of people...

You could take some inspiration from this script if you have no clue how to extract the copy_views module and rebase it on the server-tools branch https://github.com/OCA/l10n-brazil/pull/529#issuecomment-305256100


cc @aricaldeira @mileo @renatonlima @pedrobaeza

copy_views/models/inherited_ir_ui_view.py
```diff
@@ -0,0 +1,140 @@
...
+            for view_id, cv in copied_views:
+                view = self.browse(view_id)
+                arch = view.read_combined(['arch'])['arch']
+                for view in cv:
+                    view.write({'arch': arch})
+
+        return res
+
+    @api.model
+    def unlink(self):
```
renatonlima: should be @api.multi and there are other bugs, seems untested function

renatonlima (commented):
Hello guys,

I extract this module and create one PR in server-tools repo https://github.com/OCA/server-tools/pull/863

renatonlima:
As explained in OCA/server-tools#863  the feature of this module already exists in the core, using the primary attribute.

rvalyi:
related to this PR in the WIP branch: https://github.com/odoo-brazil/l10n-brazil-wip/pull/91
Here we see the benefit from interacting more with the core of the OCA to avoid re-inventing the wheel.

PR #547, opened 2017-06-09
[10.0] Merge odoo brazil wip/sped_imposto

rvalyi:


rvalyi:
@renatonlima I think we really need your review here...

rvalyi:
In term of dependency and conflict with the OCA code, I'm more optimistic about merging this module. It does depend on the WIP l10n_bar_base module which currently clashes badly with the OCA one, but most of classes only depend on `odoo.addons.l10n_br_base.constante_tributaria` and some files also depend on `odoo.addons.l10n_br_base.models.sped_base`. These classes could easily be extracted in a common l10n_br_base_lib or something similar we can share between the sped_* modules and the OCA ones.

The files that depend only on `odoo.addons.l10n_br_base.constante_tributaria`  are:

- sped_cest.py
- sped_cfop.py
- sped_cnae.py
- sped_natureza_operacao.py
- sped_nbs.py
- sped_ncm_cest.py
- sped_ncm_ibptax.py
- sped_ncm_ipi.py
- sped_ncm_protocolo.py
- sped_ncm.py
- sped_operacao_item.py
- sped_operacao.py
- sped_protocolo_icms.py
- sped_servico.py

The files that also depend on `odoo.addons.l10n_br_base.models.sped_base` are:

- sped_aliquota_icms_proprio.py
- sped_aliquota_icms_st.py
- sped_aliquota_ipi.py
- sped_aliquota_iss.py
- sped_aliquota_pis_cofins.py
- sped_aliquota_simples.py
- sped_calculo_imposto_item.py
- sped_calculo_imposto.py
- sped_ibptax.py
- sped_ibptax.py
- sped_soma_imposto.py

The 3 files which may be a bit more complex to integrate are:

- [ ] The `inherited_sped_produto.py` file depends on sped.product but it should be simple to find a solution.

- [ ] inherited_sped_participante.py depends on sped.participante but again it should be simple to find a solution.

- [ ] inherited_sped_empresa.py is more problematic as it depends on sped.empresa which is currently pretty fucked up in the WIP branch.


Now in term of conceptual model, I still have doubts how this code will play well with the Odoo account module or how it conflicts. I understand KMEE wants the sped module to be a simple electronic invoice transmissor that works without the Odoo acount module, but we should also ensure this module is compatible with the ERP usage of Odoo.

Specially, I think we should think about the integration with the following Odoo account object:
`ṕroduct.category`, `account.tax`, `account.tax.template`, `account.tax.group`, `account.account.tag`, `account.account.type`, `account.fiscal.position`, `account.fiscal.position.tax`, `account.fiscal.position.account`, `account.journal`.

So may be we should work on the field name and object inheritance model to ensure it will play well with the account module... @renatonlima, @aricaldeira, @mileo, @mbcosta what do you think?

rvalyi:
A more functional analysis now:

First I really like how you manage to factor Brazil fiscal logic between sale/purchase/invoice and their lines thanks to this module. This is really a logic we have to inject into the OCA code. 

Now in order to achieve convergence, I think we should again try to rename some WIP fields like the OCA fields first (or do the reverse if there is some big justification).

Specially we should compare fields of:

- [ ] sped_calculo_imposto.py with fields in OCA localized sale.order / account.invoice / purchase.order. See usage in https://github.com/odoo-brazil/l10n-brazil-wip/blob/10.0-develop/sped_sale/models/inherited_sale_order.py or in sped_purchase or sped_account.
- [ ] sped_calculo_imposto_item.py with fields in OCA sale.order.line / invoice.order.line / purchase.order.line. See usage in https://github.com/odoo-brazil/l10n-brazil-wip/blob/10.0-develop/sped_sale/models/inherited_sale_order_line.py or in sped_purchase or sped_account.

I think we should also analysis the impact of the WIP branch not using some core Odoo accounting objects, specially theses:

- [ ] `account.tax`
- [ ] `account.fiscal.position` (important for advanced tax and account mappings)

Can we find designs so that it works both with and without the Odoo accounting module? Like may be copy a fake/mock rudimentary tax and fiscal position objects with the same name and fields that would be properly overridden when the Odoo account module is installed? **But more importantly, is it any smart to ignore the Odoo account objects when you use the sped_imposto module in sale or purchase modules that do depend on the Odoo account module? (except the sped module)**

- [ ] Another core object of the current OCA localization is the `l10n_br_account.fiscal.category` object (can we factor the same object here?) see https://github.com/OCA/l10n-brazil/blob/10.0/l10n_br_account/models/l10n_br_account.py#L147


cc @renatonlima @mileo @mbcosta @aricaldeira

rvalyi:
a inclusao desse modulo esta reportada na branch da versao 11

PR #548, opened 2017-06-09
[10.0] Merge odoo brazil wip/sped

rvalyi:


rvalyi:
a inclusao desse modulo esta reportada na branch da versao 11

PR #550, opened 2017-06-14
Merge odoo brazil wip/financial

rvalyi:


renatonlima (changes requested):
object permission (security/ir.model.access.csv) missing

PR #551, opened 2017-06-21
Merge odoo brazil wip/l10n br resource

renatonlima:

-
-
-

--


mileo:
Vc pode fazer o rebase para corrigir o travis?

l10n_br_resource/models/resource_calendar.py
```diff
@@ -0,0 +1,303 @@
...
+    def proximo_dia_util(self, data_referencia=datetime.now()):
+        """Retornar o próximo dia util.
+        :param datetime data_referencia: Se nenhuma data referencia for passada
+                                   verifique se amanha é dia útil.
+        :return datetime Proximo dia util apartir da data referencia
+        """
+        while data_referencia:
+            if self.data_eh_dia_util(data_referencia):
+                return data_referencia
+            data_referencia += timedelta(days=1)
```
mbcosta, line 203: There is a problem here, the function return the next useful day ( proximo_dia_util ), but the first check is with Date receive, what makes wrong result, because if the date receive is useful day the function return the same date instead of the next useful day, to solve the simple way is just put "data_referencia += timedelta(days=1)" before the "if" . I saw the error by checking the tests, you can see it happening on test "def test_06_obter_proximo_dia_util(self)".

rvalyi:
guys for the record I pretend to review this. It doesn't look bad at first sight. Unfortunately it seems I will be able to do a deep review only next week...

l10n_br_resource/models/resource_calendar.py
mbcosta: I make a PR to fix this, https://github.com/odoo-brazil/l10n-brazil-wip/pull/97

rvalyi:
I reviewed the 8.0 PR here https://github.com/OCA/l10n-brazil/pull/564. @Hendrixcosta @mileo that would be cool if we can already get the 8.0 right and get the v10 version a branch of it next. Well only if this is possible, I can simply review this v10 PR once we process the v8 one. In any case if you think you cannot afford taking care of a recent PR for an older serie it is still better if you can mention it in the PR to save everybody's time.

rvalyi:
superseded by https://github.com/OCA/l10n-brazil/pull/567

PR #564, opened 2017-08-03
[NEW][8.0] Module l10n br resource

hendrixcosta:
Adicionar módulo de l10n_br_resource à versão 8.0
-
-
-

--


mileo (approved):

l10n_br_resource/models/resource_calendar.py
```diff
@@ -0,0 +1,254 @@
...
+    _inherit = 'resource.calendar'
+
+    def _compute_recursive_leaves(self, calendar):
+        res = self.env['resource.calendar.leaves']
+        res |= self.env['resource.calendar.leaves'].search([
+            ('calendar_id', '=', calendar.id)
+        ])
+        if calendar.parent_id:
+            res |= self._compute_recursive_leaves(calendar.parent_id)
+        return res
```
rvalyi, line 22: line 16 res = self.env['resource.calendar.leaves'] seems useless, line 17 would need to change too I guess. Shouldn't this be just if calender.parent_id... else... ? without these |= assignations? Or did I just miss something?

l10n_br_resource/models/resource_calendar.py
```diff
@@ -0,0 +1,254 @@
...
+    parent_id = fields.Many2one(
+        'resource.calendar',
+        string='Parent Calendar',
+        ondelete='restrict',
+        index=True)
+    child_ids = fields.One2many(
+        'resource.calendar', 'parent_id',
+        string='Child Calendar')
+    _parent_store = True
+
```
rvalyi, line 39: You may have good reasons to create a recursive calendar. But we just don't have clue yet why is this required? Could you please give an example why is this necessary (for country, state or city?) and drop an explication line about it in the module README?

rvalyi (changes requested):
First congratulation for the hard and quality work! Sorry for the delay but I could finally review this and check with @alexis-via who is quite specialized in the HR calendar/leaves management here.

Overall this seems quite right, though I have put 2 questions/comments in the code.

However we still have a doubt: you deal with public holidays entirely with the Odoo native resource.calendar.leave object which you extend consistently. Can you however understand why some people created the hr_public_holidays OCA module https://github.com/OCA/hr/blob/10.0/hr_public_holidays ? This is what @alexis-via  used so far and also extended in the Akretion HR usability module: https://github.com/akretion/odoo-usability/tree/10.0/hr_holidays_usability And it seems even another company, Onestein extend it https://github.com/onesteinbv/addons-onestein/tree/9.0/hr_public_holidays_leaves

Now may be the native resource.calendar.leaves object fits just right and these guys just miss it. What is your opinion?

Note: I also base this comment from an extensive internal review from @mbcosta 

cc @mileo @renatonlima @Hendrixcosta

rvalyi:
Hello @dreispt @jgrandguillaume,

I added you in the thread in order to get you opinion. You both are OCA/hr PSC's and we wonder if this PR is right to use the Odoo native resource.calendar.leave object ( https://github.com/odoo/odoo/blob/10.0/addons/resource/models/resource.py#L756 ) to deal with public holidays or if there is a strong justification to use instead this OCA module OCA https://github.com/OCA/hr/blob/10.0/hr_public_holidays

I'm a bit confused what the hr_public_holidays module would bring over the native resource.calendar.leave object. We got @alexis-via to have a look but he couldn't get a clue either. Is there any historical justification for the OCA hr_public_holidays module? resource.calendar.leave seems to be around since version 6.1 and hr_public_holidays is just as old so no clue here...

cc @pedrobaeza you are not HR PSC but I saw you commenting about hr_public_holidays once so may be you know the answer.

PR #542, opened 2017-06-06
[10.0] Merge odoo brazil wip/l10n br base

rvalyi:
see here for the extraction script https://github.com/OCA/l10n-brazil/pull/529#issuecomment-305256100

PR #576, opened 2017-10-04
11.0 mig sped

mileo:


rvalyi (changes requested):
Hello! thank you for the proposal and the great work packed in the tax engine. We understand the intent and want to get the tax engine merged in the OCA, however going back to the implementation detail, here is an early feedback that may help you to focus your work to get this properly integrated in the OCA (reminding you you also have pending PR's for integrating your work for boleto/CNAB and the hr_resource making the bridge with your HR incubator).

back on v10 I made these reviews:
https://github.com/OCA/l10n-brazil/pull/529
https://github.com/OCA/l10n-brazil/pull/548
https://github.com/OCA/l10n-brazil/pull/547
https://github.com/OCA/l10n-brazil/pull/544
https://github.com/OCA/l10n-brazil/pull/542

Most of problems still persist for integration with the OCA localization. And in any case you provided NO feedback to these reviews. Frankly as the early design with the inherits everywhere and no test was really clumsy, we just thought you abandoned the idea and where surprised to see making these PR in v11 again without answering us on the v10 even. Which is sad because I think you did a lot of absolutely useless work, probably the majority of the code here, but I will prove that to you in a few days when I will have more time.

At that time me and @mbcosta did a very deep systematical analysis of the fields you are using and explicited the diff in some tables and asked you to use the OCA field names.

Well, in fact your posposal forced us to dig into this issue to deal better with the fiscal information of Brazilian Invoices (for the SPED mostly).

We are still focusing our efforts on getting the localization migrated, much like Trustcode did so we can then evalute better the integration of your work.

So it is a bit early for us, but so far I can already tell that, provided you can exibit extensive tests, we are extremely interrested in your tax engine, that is this file:
https://github.com/kmee/l10n-brazil/blob/11.0-develop/sped_imposto/models/sped_calculo_imposto_item.py

Eventually this could be used as a fall back when no tax rule is found (and could be used extensively for the companies under the SIMPLES fiscal regime if I make no mistake).

To be integrated in the OCA we would need tests for sure and we asked you change the fields name in that review 5 month ago and during the early sprints.

Today I changed my mind, I think we will also change the field names of all the fields that go in the electronic invoice in the OCA localization and be systematic about that (but not change the fields and tables and codebase in general except these specific fields for which the migration may be trivial).

So you cannot refactor your code reliably if there is no unit test FIRST. And this is the reason why @renatonlima @mbcosta and me did not do this refactor ourselves as we asked you to provide an extensive test coverage first.

I would say, to get this code integrated into the OCA, focus on testing this file specially:
https://github.com/kmee/l10n-brazil/blob/11.0-develop/sped_imposto/models/sped_calculo_imposto_item.py

Then we will make a proposal for the storage and handling of the fiscal information that will not break a single test in the OCA localization (unlink your rewrite approach that we cannot accept). So once you have this part of the code tests, if you cannot work on this further, Akretion makes the commitment to adapt the code and field naming so we can integrate this into the OCA tax engine. I think the potential is huge.

You may feel dubitative about this field renaming we will propose for the OCA codebase. But this is because you didn't see our proposal yet. In fact I started to work on this only 2 weeks ago and soon I could see that what you propose can be done in a much more reliable way without breaking 7 years of battle tested operational integration of the current OCA localization (and without saying fuck you to all the customers who invested some faith into the OCA stability promise). This is too early for me to expose this proposal yet and I prefer helping @renatonlima getting the localization fully migrated as I we think this is the very first step to be done before talking integration of new features with such a deep impact.

cc @mileo @renatonlima @mbcosta 

Regards.

PR #581, opened 2017-10-17
8.0 l10n br account banking payment

rvalyi:


rvalyi:
gah, it looks like I screw the git-filter operation and commits where not filtered properly. I'll fix this soon sorry.

PR #591, opened 2017-11-15
[FIX] fix unicode string in l10n_br_base python tests

clementmbr:
Fix from akretion/10-convert-test-clement


renatonlima (approved):

rvalyi (approved):
I still don't get why the test refactor branch was green before the merge and not after it was merged, but these changes are correct so this may fix the problem.

mbcosta (approved):

PR #593, opened 2017-12-08
attempt to fix unicode issue

rvalyi:
The errors we had in Travis https://travis-ci.org/OCA/l10n-brazil/jobs/303167245 was a bit mysterious as PR tests were passing but not the test in 10.0 branch. So this is a bit hard to investigate but here is an attempt to fix the issue that is caused by a "ç" char as one ca see doing ```print u'\xe7'```. cc @clementmbr

### DIFF
--- a/l10n_br_base/tests/test_valid_createid.py
+++ b/l10n_br_base/tests/test_valid_createid.py
@@ -111,40 +111,40 @@ class ValidCreateIdTest(TransactionCase):
 # Tests on companies
 
     def test_comp_valid(self):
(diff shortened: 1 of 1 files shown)


mbcosta (approved):

rvalyi:
note: this is probably not the cleanest fix. But the point is we don't care to have c or ç in a test assert or docstring and because these tests use to pass in the PR and not once it is merged, we have to take the whole OCA PR/merge dance just to find out if the fix works, this is why I propose this quite brutal fix.

renatonlima (approved):

PR #623, opened 2018-08-23
10.0 mig l10n br sale

mileo:
Depende de #603


rvalyi (changes requested):
same comment:
thanks!! however you need to trim the ```<data></data>``` tags as they are not required anymore. Eventually as there is no XML change, you can also make a separate commit to ensure XML indentation is conform to OCA standards if it is not (the sooner we tackle these things, the earlier we make cherry picking between next branches easy).

rvalyi:
can you rebase on 10.0 and fix the conflicts please?

rvalyi (dismissed):
Thanks for the work. However, the module has not been set as uninstallable in the manifest file. That is why the coverage didn't decrease. We would need tests too... Typically the test from l10n_br_sale_stock could be adapted. But that would be good to have the test in Python rather than yaml then... cc @renatonlima @mbcosta

the-bmb:
Depende do PR https://github.com/OCA/account-fiscal-rule/pull/86

mileo:
Substituído por https://github.com/OCA/l10n-brazil/pull/648

PR #624, opened 2018-08-23
10.0 mig l10n br sale product

mileo:
Depende de #623 #602


rvalyi (changes requested):
thanks!! however you need to trim the ```<data></data>``` tags as they are not required anymore. Eventually as there is no XML change, you can also make a separate commit to ensure XML indentation is conform to OCA standards if it is not (the sooner we tackle these things, the earlier we make cherry picking between next branches easy).

rvalyi:
closing this one as it is superseded by https://github.com/OCA/l10n-brazil/pull/710 which was merged.

PR #626, opened 2018-08-23
RUNBOT TEST - DO NOT MERGE

rvalyi:


rvalyi:
this is just a demo of https://github.com/OCA/l10n-brazil/pull/627 but with an existing PR. The Runbot is up.

PR #627, opened 2018-08-23
ressucitate Runbot by removing pysped and upgrading Travis conf

rvalyi:


rvalyi:
so as mentioned in the chat, the only way to have Runbot running in our project is by removing the pypsed dependencies. Without pysped, Runbot is up in 9 minutes. With it, it timeouts at 25 minutes.

This is pretty much in line with the design of the OCA/l10n-brazil repo and the https://github.com/odoo-brazil/odoo-brazil-eletronic-documents repo that should bare with the pysped dependencies it assumes.

So for now I remove pysped from the repo. We only needed it here:
https://github.com/OCA/l10n-brazil/blob/8.0/l10n_br_account_product/sped/nfe/document.py

I propose we move this file into the https://github.com/odoo-brazil/odoo-brazil-eletronic-documents repo in v10 and do further refactoring to allow substituting pypsed by something more maintainable, possibly https://github.com/danimaribeiro/PyTrustNFe and eventually nfelib.

I left pybrasil in the requirements.txt, but that will be up for debate if we don't want pybrasil to become the new pysped in term of untested monolithic code that is unreasonable to pretend maintain and can defeat the whole promise of maintenable code in the OCA if modules depend on it.

renatonlima (approved):
I agree! In the future we will remove the dependency of the pysped.

mbcosta (approved):

PR #643, opened 2018-08-29
[8.0][FIX] Fix import and field name of move line

mbcosta:
Fixed wrong import and field 'name' of account.move.line because it is used on implementation of "Boleto", before when the field 'Referência/Descrição' is filled this value was used in the name


renatonlima (approved):
#TODO cherry-pick these commits in  10.0 branch.

mileo (approved):

PR #652, opened 2018-12-28
[12.0] mig l10n br base

renatonlima:
Migrate to 12.0


BiancaBartolomei:
Instalei o módulo e testei também, aparentemente tá tudo certo. Tem só alguns detalhes pra ficar certinho, que é no arquivo sped_account_payment_term_line_view tem um <label> sem for, no arquivo sped_produto_base_view tem um <button> sem string assim como no sped_empresa_vincula_company_view e existe o fields.Date.from_string que não tem mais necessidade nos arquivos inherited_acount_payment_term e copied_account_payment_term. Arrumando isso, já aprovo.

BiancaBartolomei (approved):

rvalyi:
Folks, I worked hard on the fiscal code generation with generateDS these days and started the tests on v12. It is not so easy to share the code as I am ensuring that is work for all the fiscal documents, both for the XML parsers/generators lib and the Odoo mixins. And also I'm still doing adjustments on the way to organize the branches of the code generator and generated libs/modules to ensure it is easy to track the changes yet separating the manual changes from the code generation. So that's why.

But please don't merge this PR yet before my OK. I tested the module, the code looks great (kudos @renatonlima !!) but I will still check there is no catch when we will integrated with the generateDS fiscal mixins in the fiscal modules that will depend on it. The idea is this module should still have the hand written code for the fiscal fields that will be common to the next fiscal modules like nfe, nfse, mdfe, cte... So in theory we will still keep that module this way, but still I need to validate a few things and then we merge it. Thank you all for your work and review, this is very much appreciated! 

BTW I now master every option of generateDS and I can tell you it is going to rock!!

rvalyi:
All right folks! Tested with the generateDS mixins and we will be all good! The only thing we could later debate is if it would be interesting to move some legal field to later modules such as fiscal or l10n_br_account but as we had them already I see no trouble in having them all as a start, specially as they are not required here, which doesn't harm multi-country setup.

PR #667, opened 2019-03-31
[REM] Unicode

hendrixcosta:

-
-
-

--

### DIFF
--- a/l10n_br_base/models/res_company.py
+++ b/l10n_br_base/models/res_company.py
@@ -118,7 +118,7 @@ class Company(models.Model):
         ondelete='cascade')
 
     inscr_mun = fields.Char(
-        string=u'Municipal Tax Number',
+        string='Municipal Tax Number',
         compute='_compute_l10n_br_data',
         inverse='_inverse_inscr_mun',
         size=18)
--- a/l10n_br_base/models/res_config_settings.py
+++ b/l10n_br_base/models/res_config_settings.py
@@ -8,7 +8,7 @@ class ResConfigSettings(models.TransientModel):
     _inherit = 'res.config.settings'
 
     allow_cnpj_multi_ie = fields.Boolean(
-        string=u'Multiplos parceiros com mesmo CNPJ',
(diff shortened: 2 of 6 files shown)


hendrixcosta:
@rvalyi

rvalyi (approved):

rvalyi:
@hendrixcosta thanks!
BTW, @renatonlima fixed the tests for l10n_br_account_product on v10, he just need to push the changes. The port fro v10 to v12 of the accounting module is pretty trivial. But as discussed with @mileo we did a big refactor that consists mostly in:

1. extract a fiscal module both from l10n_br_account and l10n_br_account_product a bit like the sped/sped_imposto module from Ary, but lighter, fully compatibly with Odoo accounting and the OCA localization. This is nearly done, you should see the PRs landing in the very next days.
2. Also we don't try to maintain all the fiscal data manually like Ary did. And mostly, we think it was a totally crazy idea to cripple the ERP operations tables with fiscal data from the SPED consolidation reports spec that change every year and try to fit both the SPED reports spec and the fiscal documents specs with the same data model as the specs update often and with a DIFFERENT calendar! That means you would need complex ERP migrations like OpenUpgrade several times a year, even on very small companies with no budget that don't care at all about SPED reports... No OCA localization would ever attempt such a design... So instead, we code generated the mixins of the fiscal documents on one hand and of the SPED reports on the other hand. The PR's for the fiscal documents mixins are not 100% finished yet but they are already available for your early evaluation. So on my side, I'm prototyping the new l10n_br_account_product module that will used the NFe mixins and get all the XML marshalling/unmarshalling, data validation and many forms all automatically maching exactly the fiscal specification. Then that technology will be marged with the new fiscal module and other refactors @renatonlima is doing in the l10n_br_account module. Other such fiscal modules using the generated mixins could then be contributed to complete the fiscal documents managements. Same for the SPED reports then.

So, thanks for you help and don't desperate, great stuff is coming from our side too and the v12 OCA localization will kill it all!

mileo (approved):

PR #682, opened 2019-05-07
[10.0][FIX] BIC Field

mbcosta:
There was a problem with the field "bic" from object "res.bank" it was used to put another information in our localization, brazilian code bank, I openned a issue about https://github.com/OCA/l10n-brazil/issues/588 , in this PR I purpose a solution for the problem.


renatonlima (approved):

rvalyi (approved):
OK, esta conforme ao que foi conversado um tempo atras. Valeu.

PR #689, opened 2019-05-15
[FIX] Travis error of l10n_br_resource

mileo:
Signed-off-by: Luis Felipe Mileo <mileo@kmee.com.br>

### DIFF
--- a/l10n_br_resource/tests/test_resource_calendar.py
+++ b/l10n_br_resource/tests/test_resource_calendar.py
@@ -2,7 +2,6 @@
 # Copyright 2016 KMEE - Luis Felipe Mileo <mileo@kmee.com.br>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-
 from odoo import fields
 import odoo.tests.common as test_common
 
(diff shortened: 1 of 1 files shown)


renatonlima (approved):

hendrixcosta (approved):

rvalyi (changes requested):
It looks like Travis is unhappy https://travis-ci.org/OCA/l10n-brazil/jobs/532887301
I'm not too sure, but it looks it's because of the ```unicode_literals``` import that some str objects are changed into unicode and makes the test runner unhappy.
```
File "/usr/lib/python2.7/unittest/runner.py", line 44, in getDescription
doc_first_line = test.shortDescription()
File "/home/travis/odoo-10.0/odoo/tests/common.py", line 135, in shortDescription
return doc and ' '.join(filter(None, map(str.strip, doc.splitlines()))) or None
TypeError: descriptor 'strip' requires a 'str' object but received a 'unicode'
```

hendrixcosta (approved):

rvalyi (approved):

renatonlima (approved):

PR #707, opened 2019-08-20
[10.0] [FIX] Error postgres and remove deprecated Transifex in Travis 

mbcosta:
Update Postgres configuration on Travis and remove deprecated Transifex.

cc @renatonlima @rvalyi @mileo

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
   - "2.7"
 
 addons:
+  postgresql: "9.6"
   apt:
     packages:
       - expect-dev  # provides unbuffer utility
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ addons:
 env:
   global:
   - VERSION="10.0" TESTS="0" LINT_CHECK="0"
-  - WKHTMLTOPDF_VERSION="0.12.4"
+  - WKHTMLTOPDF_VERSION="0.12.5"
 
   matrix:
   - LINT_CHECK="1"
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,16 +17,13 @@ addons:
 
 env:
   global:
-  - VERSION="10.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0"
-  - TRANSIFEX_USER='transbot@odoo-community.org'
+  - VERSION="10.0" TESTS="0" LINT_CHECK="0"
   - WKHTMLTOPDF_VERSION="0.12.4"
-  - secure: ArnbVaF5+ry6zVysZ7HA9xcnQrodo8FsXXcon9yINTYRfDC9QEr+/mTLAPRl+lVLdtWV2GuGuX0vPYzBxFWpY3LR0BpKqXzx0G51s94zR2WWEmizYFzFhpAuIxoU4CYNckHFSUaPAQJhwB/pYx9+H/W6bMjG/VnZBq+AmBJ2Kh0=
 
   matrix:
   - LINT_CHECK="1"
-  - TESTS="1" ODOO_REPO="odoo/odoo"
+  - TESTS="1" ODOO_REPO="odoo/odoo" MAKEPOT="1"
   - TESTS="1" ODOO_REPO="OCA/OCB"
-  - TRANSIFEX="0"
 
 virtualenv:
   system_site_packages: true


.travis.yml
```diff
@@ -16,14 +17,11 @@ addons:
 
 env:
   global:
-  - VERSION="10.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0"
-  - TRANSIFEX_USER='transbot@odoo-community.org'
+  - VERSION="10.0" TESTS="0" LINT_CHECK="0"
```
rvalyi, line 20: I'm not too sure if it has a real impact, but in the template the line should include ```MAKEPOT="0"```
I guess this would avoid useless .pot generation (possibly we want to generate it only once).
see https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.travis.yml#L32

.travis.yml
```diff
@@ -16,14 +17,11 @@ addons:
...
   - WKHTMLTOPDF_VERSION="0.12.4"
```
rvalyi: let's move it to 0.12.5 as suggested https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.travis.yml#L36

.travis.yml
```diff
@@ -16,14 +17,11 @@ addons:
...
-  - secure: ArnbVaF5+ry6zVysZ7HA9xcnQrodo8FsXXcon9yINTYRfDC9QEr+/mTLAPRl+lVLdtWV2GuGuX0vPYzBxFWpY3LR0BpKqXzx0G51s94zR2WWEmizYFzFhpAuIxoU4CYNckHFSUaPAQJhwB/pYx9+H/W6bMjG/VnZBq+AmBJ2Kh0=
-
-  matrix:
+   matrix:
   - LINT_CHECK="1"
-  - TESTS="1" ODOO_REPO="odoo/odoo"
+  - TESTS="1" ODOO_REPO="odoo/odoo" MAKEPOT="1"
   - TESTS="1" ODOO_REPO="OCA/OCB"
   - TRANSIFEX="0"
```
rvalyi: this TRANSIFEX line should be removed

.travis.yml
```diff
@@ -16,14 +17,11 @@ addons:
...
-  - TRANSIFEX_USER='transbot@odoo-community.org'
+  - VERSION="10.0" TESTS="0" LINT_CHECK="0"
   - WKHTMLTOPDF_VERSION="0.12.4"
-  - secure: ArnbVaF5+ry6zVysZ7HA9xcnQrodo8FsXXcon9yINTYRfDC9QEr+/mTLAPRl+lVLdtWV2GuGuX0vPYzBxFWpY3LR0BpKqXzx0G51s94zR2WWEmizYFzFhpAuIxoU4CYNckHFSUaPAQJhwB/pYx9+H/W6bMjG/VnZBq+AmBJ2Kh0=
-
-  matrix:
+   matrix:
   - LINT_CHECK="1"
-  - TESTS="1" ODOO_REPO="odoo/odoo"
+  - TESTS="1" ODOO_REPO="odoo/odoo" MAKEPOT="1"
```
rvalyi, line 25: @renatonlima if we add this MAKEPOT="1", I believe .pot files will be generated and possibly .po files might be overriten by the weblate translation service. I believe that starting from v12 we will be quite compatible to use the Brazilian localization along with other localizations, so activating the weblate translation makes sense. I just fear it may override the existing .po files of the project. Really not sure about that. What do you think?

rvalyi (changes requested):
thanks. minor changes requested

.travis.yml
renatonlima: @rvalyi,

In l10n-brazil we always create fields and objects names in english, but all labels are in portuguese, now in version 12.0 we are putting all labels also in English, For example in module l10n_br_base I generated a POT file with the module terms and also uploaded one file pt_BR.po already with the portuguese translations, weblate synchronized the translations and didn't overwrite it so I would say there is no problem.

rvalyi: indeed we should have no problem https://twitter.com/SBidoul/status/1163876494105612288

rvalyi (approved):

.travis.yml
renatonlima: 👍

renatonlima (approved):

rvalyi:
gah... the .pot generation just failed on 10.0
https://travis-ci.org/OCA/l10n-brazil/jobs/574540770#L1841

I don't know why yet. Also to avoid such errors in the future, instead of using the merge button, we should use the /ocabot merge [major|minor|patch] bot directive that will re-run the tests as it will be done when merged so such errors would be detected and the merge aborted. Now let's fix that one...

PR #708, opened 2019-08-29
[10.0][ADD] pot translation files

renatonlima:


rvalyi (approved):

mbcosta (approved):

rvalyi:
@renatonlima there still an error https://travis-ci.org/OCA/l10n-brazil/jobs/578858740#L1912
At least that ocabot workflow is able to detect these .pot errors that we don't spot ourselves.

PR #710, opened 2019-09-02
[10.0][MIG] l10n_br_sale_product

mbcosta:
Replace https://github.com/OCA/l10n-brazil/pull/624 to use a common repository.

cc @renatonlima @rvalyi @mileo


l10n_br_sale_product/models/sale_line.py
```diff
@@ -0,0 +1,95 @@
+# -*- coding: utf-8 -*-
+# @ 2019 Akretion - www.akretion.com.br -
+#   Magno Costa <magno.costa@akretion.com.br>
```
rvalyi: as this file is largely copied from sale.py, it should also mention
Copyright (C) 2014  Renato Lima - Akretion

renatonlima (approved):

mbcosta:
Thanks @renatonlima and @rvalyi for review

I made the changes you ask for

l10n_br_sale_product/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module extends the standard Odoo sale module for Brazil. Especially it deals with the propagation of the fiscal operation and fiscal comment. This module can be used both with or without the "NFe" electronic invoicing so it does not provide the NFe support itself.
```
rvalyi: sorry @mbcosta I made a mistake: l10n_br_sale_product is specially designed for sales with NFe. So please chane it to just:

This module extends the standard Odoo sale module for Brazil, specially when selling products with NFe's.

It deals with the propagation of the fiscal operation and fiscal comment. I also displays the following sale total details:

* untaxed amount
* taxes
* gross amount (valor bruto)
* shipping (frete)
* costs (outros custos)
* insurance (seguro)

l10n_br_sale_product/models/sale.py
```diff
@@ -134,65 +133,91 @@ def _set_amount_costs(self):
...
+        compute='_amount_all_wrapper',
+        string='Untaxed Amount',
+        digits=dp.get_precision('Account'),
+        store=True,
+        help="The amount without tax.",
+        track_visibility='always')
     amount_tax = fields.Float(
-        compute='_amount_all_wrapper', string='Taxes', store=True,
-        digits=dp.get_precision('Account'), help="The tax amount.")
+        compute='_amount_all_wrapper',
```
rvalyi: back to v7 or v8, because of the old API, that _amount_all_wrapper wrapper was needed as you can see here:
https://github.com/odoo/odoo/blob/8.0/addons/sale/sale.py#L59
However in v10, with the genaralized API this is not required anymore as you can see here:
https://github.com/odoo/odoo/blob/10.0/addons/sale/models/sale.py#L22 (no more _amount_all_wrapper method)

So you could please delete this _amount_all_wrapper and use _amount_all directly?
Also at the beginning of the _amount_all metho override, please add the following comment:

"This override is specific for Brazil. Idealy, starting from v12 we should just call super when the sale is not for Brazil".

(This multi-localization ability is very much required. However, Akretion won't spend efforts in doing them for v10).

rvalyi (changes requested):
Thank you @mbcosta for the good work! However I requested some simple corretions. Then I think we will be able to merge it.

mbcosta:
Hi @rvalyi thanks for review, I made the changes

rvalyi (approved):
thanks!

rvalyi:
And so we have an issue on the .po generation https://travis-ci.org/OCA/l10n-brazil/jobs/583366516#L2035
BTW I'm not sure this was so smart to enable weblate in 10.0 as the focus is now on 12.0 and we lived without till then. On the other hand v10 is nearly migrated so eventually we deal with making Weblate happy...

PR #658, opened 2019-03-21
[12.0][ADD] l10n_br_spec_nfe

rvalyi:
This a work in progress of the NFe mixins matching the XSD spec and generated with the Odoo generateDS plugin. NOT READY FOR MERGE YET.

PR #716, opened 2019-09-14
[12.0][ADD] l10n_br_nfe_spec - WIP

rvalyi:
attempt to supersede https://github.com/OCA/l10n-brazil/pull/658
with NFe XML importation into Odoo mixins tests.
WIP


rvalyi:
important note:
all fiscal *_spec modules generated with the Odoo generateDS plugin should share the same AbstractSpecMixin model. However it might be annoying to make these models all depend on a new common module or depend on l10n_br_fiscal which is overkill.
So see if we move AbstractSpecMixin to l10n_br_base or if we find a way to load the complex import able AbstractSpecMixin only during the XML input/output tests.

marcelsavegnago:
@rvalyi essa PR acho que pode ser fechada. O que acha ?

rvalyi:
superseded by https://github.com/OCA/l10n-brazil/pull/981

PR #726, opened 2019-10-01
[FIX] runbot

gurneyalex:
the runbot command would hang at a password prompt caused by sudo. 

Decision taken to remove the tweaking of the locales in 12.0 by @rvalyi

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,11 +31,7 @@ install:
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
   - travis_install_nightly
 
-before_script:
-  - chmod +x configure_locale.sh
-
 script:
-  - sudo ./configure_locale.sh
   - travis_run_tests
 
 after_success:
--- a/configure_locale.sh
+++ b/configure_locale.sh
@@ -1,4 +1,0 @@
-#!/bin/bash
-
-sudo locale-gen en_US en_US.UTF-8 pt_BR.UTF-8
-sudo dpkg-reconfigure locales


ananiasfilho:
@gurneyalex can i help you on l10n-br  deployment on runbot?
if yes, let me know where you get issues from configure_locales.sh under build deployment (CI), just remove these lines from configure_locales.sh:
```
sudo locale-gen en_US en_US.UTF-8 pt_BR.UTF-8
sudo dpkg-reconfigure locales
```

But, to deploy l10n-br, you need locales on machine/instance/container/vm that will run odoo.
What we does here:
run commands before deploy odoo/runboot on local instance/vm/machine/container
```
sudo locale-gen en_US en_US.UTF-8 pt_BR.UTF-8
sudo dpkg-reconfigure locales
```
after ir, you can deploy environment.
Its help? or you have more informations so i can help you and try here.

rvalyi:
Hello all, I propose to simply remove the pt_BR system locale. This was added to enable the legacy pysped library on versions previous to v12. But starting from v12 we have a tested alternative to pysped using akretion/nfelib (geneared code via generateDS) so I propose we just drop these lines. Feel free to make the PR, otherwise I'll be doing it.

That won't fix the runbot on v10 though. We have two easy alternative on v10: drop pysped as well, but the test coverage would decrease a lot. Or live without Runbot on v10. I would say live without runbot on v10. It's not the focus anymore. With focused work we can soon finish to port things to v12....

ananiasfilho:
@rvalyi i agree with v12, but with v10 and v8, i think we need make these changes to run on runbot. A lot of users are using v10 and v8 from OCA Localization and we are responsable about it, agree?

rvalyi:
@ananiasfilho 
So yes, let's drop pt_BR on v12.

As for v8 and v10 that would be cool we find a workaround indeed. Something interresting is also to be able to quickly compare the features when we will be porting the modules to v12. We the lead developpers install everything, but not the occasional reviewer.

Something that may be responsible for the failure might be timeout. In my experience compiling a new locale easily adds 4 or 5 minutes to the build. And Runbot has a timeout. I'm not sure if it was 15 or 20 minutes, but may be we are reaching it because of this extra time. Then if somebody finds a way to have a custom image or have pysped don't complain about the locale if it's not there... (we don't really care to have compliant NFe's on the v8 or v10 Runbot but we don't want to have Travis tests failing either).

As for if we are "responsible" or not, I agree more or less, people should not believe there are free lunches with anybody guarantying them anything more than enforcing the debated OCA admission criteria. We already tend to have a very amateurish community over there (and I think this time Luis had an opportunity to see how different it is in Europe where Odoo is strong), no need to keep them dreaming that OCA is an Odoo distro o things like that (it could soon be, but certainly not on v8 or v10). IMHO it's way more important to make sure our community understands well they better contract with OCA contributors to ensure some work is put in the project rather than the usual middlemen who use to run away with their money and leave them with a broken ERP.

gurneyalex:
@rvalyi I'm removing the locales tweaks in this PR and in parallel I reverted the attempts to fix I'd made on runbot.

gurneyalex:
runbot is cool for l10n-brazil 12.0 with this PR: 
image

rvalyi:
Thank you @gurneyalex . So now this looks very much like my other PR https://github.com/OCA/l10n-brazil/pull/729/files where I even go further into prunning unused legacy dependencies. But we can merge this one no problem for me.

rvalyi (approved):

gurneyalex:
@rvalyi note that with the runbot docker image tweak remove #729 build is green too on runbot. So feel free to merge #729 instead of this one, esp. if it is more complete. 

I apologize for forgetting this after getting back from Louvain la Neuve. I was really tired and mostly slept through the week end...

rvalyi:
@gurneyalex well yes in fact I think using only the required dependencies is better... I re-openned #729, let's see with the others.

mileo (approved):

PR #731, opened 2019-10-08
[12.0][WIP] l10n_br_account 10n_br_fiscal bridge

rvalyi:
This is the 1st step for preparing my work with the generateDS XSD spec driven fiscal models for Odoo... However, this foundation work leaves the l10n_br_fiscal and l10n_br_account agnostic from my generateDS work (next PRs with the NFe modules coming soon). So if you prefer to manage and maintain fiscal data manually and use pysped or PyTrustNFe, then you can still team on these core OCA modules.

In this PR, we extend the work from @renatonlima on l10n_br_fiscal from https://github.com/OCA/l10n-brazil/pull/657 (which is a clean and compatible refactor l10n_br_account_product along with new fiscal concepts proposed in v11 by Aristides) and bridge it with l10n_br_account which is the base foundation both for NFe, NFSe and other fiscal documents you may want to handle through Odoo invoices.

This PR is not meant to be merged as it is. It is meant for test and evaluation instead. Eventually it will be merged inside the branch of the l10n_br_fiscal extraction here https://github.com/OCA/l10n-brazil/pull/657 instead. ONLY MY OWN EXTRA COMMITS SHOULD BE EVALUATED HERE.

Basically, l10n_br_fiscal introduces an independent concept of l10n_br_fiscal.document (and lines) which is designed to serve as the basis of all Brazilian fiscal documents even without depending on the Odoo account module (so you can have simple and easy to maintain electronic document modules if you wish; this requirement was introduced by KMEE in v10/v11).

However, the largest use case will be to handle these fiscal documents through the Odoo account.invoice object (and account.move in v13), just like we did from Odoo 5.0 to Odoo 10.0. For instance create a NFe or a NFSe via an Odoo invoice. Then we also expect that the user will use the Odoo invoice interface to edit the fiscal document data instead of having to browse different windows. For instance the product decsription, price and quantity should be just the ones of Odoo invoice lines.

This is done with _inherits:
```
class AccountInvoice(models.Model):
    _inherit = 'account.invoice'
    _order = 'date_invoice DESC, number DESC'

    _inherits = {'l10n_br_fiscal.document': 'fiscal_document_id'}
```

```
class AccountInvoiceLine(models.Model):
    _inherit = 'account.invoice.line'
    _inherits = {'l10n_br_fiscal.document.line': 'fiscal_document_line_id'}
```

However, we should make a few more considerations:

A) the case of demo invoices created without the l10n_br_fiscal module and also the case where you use Odoo in a multinational company where some invoices are not submitted to the Brazilian localization.

This is why we created a dummy electronic document with an active=False filter to make the ORM happy in these cases. And we also introduced a hook which will properly patch the existing invoices to get them use that dummy fiscal document.

With the same reasoning, some draft invoice may not have fields that are usually mandatory in Brazilian fiscal documents. This is the reason I removed the required=True attribute from these fields and ask for a softer validation through Python constraints and views.

B) the colliding fields. Befause you can both use fiscal document independently or handle them through other Odoo objects (like invoice) through _inherits, you may have fields collision. For instance the fiscal doc line amount with the invoice line amount.
It is possible to rename the fiscal document fields to anticipate and avoid such collision (like have a field fiscal_doc_line_amount).
However this is not great because:

1. it would prevent to use the same mixins in a fiscal document and the Odoo object
2. it would make much longer fields names everywhere and make code harder to read
3. for some fields it is not even possible. For instance, the fiscal document state can only be named state. Also currency_id would it cumbersome to rename and use a currency_field_name everywhere...

For all these reasons, we prefered instead to declare a few related fields aliases for all fields that collide with the invoice fields. Then inside an invoice form or method you can write the fiscal document fields using these aliases.

Yes these 14 collinding fields represent DATA DUPLICATION. However, as long as you want to be able to have independent fiscal documents and account invoices objects there isn't much choice. In fact the real benefit of this technical solution is we duplicate (and in a very systematic process) ONLY 14 fields among the 800 fiscal fields NFe's have... This is a much better situation than the Trustcode fork for instance where they duplicate every fiscal field they want to deal with in invoices...

cc @renatonlima @mileo @mbcosta


rvalyi:
just rebased so it can be tested on Runbot. Although it's more @renatonlima work that can be tested via the UI. In any case, you can still ensure my _inherits bridge has no side effect in the non NFe usage of account.invoice, which is an essential feature of the new design.

rvalyi:
@hendrixcosta ja que vc tem experiencia e esta de review, seria bom vc começar a olhar essa extração do modulo fiscal l10n_br_fiscal e a forma como se articula com o account.invoice no contexto da NF-e ou NFS-e nesse PR. Ainda devo subir o PR com a NFe muito mais completa, mas já entender a arquitetura basica eh importante, ainda mais que ate ai isso eh agnóstico do generateDS.  (mas eu acho que ja consegui comecar a convencer o Luis a usar, porque nas libs de transmissão ele já comecou a refatorar para usar e se livrar do pySPED). Da para testar o l10n_br_fiscal no Runbot apesar de estar vermelho.

l10n_br_account/models/account_fiscal_position_abstract.py
```diff
@@ -0,0 +1,36 @@
+# Copyright (C) 2009 - TODAY Renato Lima - Akretion
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from odoo import fields, api
+
+
+class AccountFiscalPositionAbstract(object):
```
hendrixcosta, line 7: Classe inutilizada

l10n_br_account/demo/base_demo.xml
```diff
@@ -0,0 +1,180 @@
+<?xml version="1.0" encoding="utf-8"?>
```
hendrixcosta, line 1: Nomenclatura não deveria ser res_partner_demo?

l10n_br_fiscal/models/account_fiscal_position.py
```diff
@@ -0,0 +1,185 @@
...
+            if mapping[tax].get('tax'):
+                result |= mapping[tax].get('tax')
+        return result
+
+    @api.model
+    def map_tax_code(self, taxes, product=None, partner=None):
+        return self._map_tax(taxes, product, partner)
+
+
+class AccountFiscalPositionTax(models.Model):
```
hendrixcosta, line 144: Não deveria ficar em arquivo separado?
 (tem boa pratica pra isso? quando o modelo tiver mais do que 4 campos

l10n_br_fiscal/models/account_payment_term.py
```diff
@@ -0,0 +1,16 @@
...
+
+class AccountPaymentTerm(models.Model):
+    _inherit = 'account.payment.term'
+
+    indPag = fields.Selection(
+        selection=[('0', u'Pagamento à Vista'),
+                   ('1', u'Pagamento à Prazo'),
+                   ('2', 'Outros')],
+        string=u'Indicador de Pagamento',
+        default='1'
```
hendrixcosta, line 15: default='1' ? Não seria melhor igual a '0' ?
Por padrão pagamento a vista e em dinheiro - caso mais simples

l10n_br_fiscal/models/certificate.py
```diff
@@ -0,0 +1,111 @@
...
+        string='Password')
+
+    is_valid = fields.Boolean(
+        compute='_compute_is_valid',
+        string='Is Valid?',
+        store=False)
+
+    @api.multi
+    @api.constrains('file', 'password')
+    def _check_certificate(self):
```
hendrixcosta, line 67: Se não me engano já tem função para teste básico de certificado.
Poderia inclusive validar a data de expiração
Ou se não remover essa constrains

l10n_br_fiscal/models/comment.py
```diff
@@ -0,0 +1,66 @@
...
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from odoo import models, fields, api
+from odoo.osv import expression
+
+from ..constants.fiscal import COMMENT_TYPE, COMMENT_TYPE_DEFAULT
+
+
+class Comment(models.Model):
+    _name = 'l10n_br_fiscal.comment'
```
hendrixcosta, line 11: Não entendi muito bem esse cara, mas não daria pra usar as propriedades de mail.thread nativa?

hendrixcosta (changes requested):

l10n_br_account/models/account_fiscal_position_abstract.py
rvalyi: ola @hendrixcosta valeu pelo retorno. Porem, o que vc comentou se refere tudo ao modulo l10_n_br_fiscal no qualo o @renatonlima ainda esta trabalhando, seria melhor deixar no PR original da extração do l10n_br_fiscal aqui https://github.com/OCA/l10n-brazil/pull/657 Aqui eu submeti apenas para avaliacao a forma de como o documento fiscal generico se atrela ao account.invoice (e account.move na 13.0). Mas mesmo assim, vamos analisar o que vc reportou.

rvalyi:
My 3 commits for the PR have been fully cherrypicked in the l10n_br_fiscal and l10n_br_account PR here https://github.com/OCA/l10n-brazil/pull/657 so I'm closing this PR.

PR #729, opened 2019-10-08
disable deprecated pysped requirements to fix Runbot

rvalyi:
supersedes #726

PR #733, opened 2019-10-09
[REF] .travis.yml: Installing language-pack-br in order to avoid using "sudo"

moylop260:
Thread about: https://github.com/Vauxoo/travis2docker/pull/126#issuecomment-539769969

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ addons:
       - libxml2-dev
       - libxmlsec1-dev
       - libxslt-dev
+      - language-pack-br
 
 env:
   global:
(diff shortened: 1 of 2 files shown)


rvalyi:
@moylop260 thank you very much for your support!! I promiss we will soon test this on the v8 and v10 branch (we don't use that anymore in v12, in fact we will just drop the configure_locale thing in v12). If it works getting the Runbot back is really great news.

rvalyi (changes requested):
Better not merge that on 12.0 and merge https://github.com/OCA/l10n-brazil/pull/729 instead as it won't bloat the tests with useless dependencies. But we should definitely try the same trick on the 10.0 branch and eventually 8.0.

moylop260:
I got it @rvalyi 
Regards!

PR #745, opened 2019-10-20
10.0-travis-locale-fix

rvalyi:
should fix Runbot on 10.0 by avoiding sudo.
Equivalent of https://github.com/OCA/l10n-brazil/pull/733 but for 10.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ addons:
     packages:
       - expect-dev  # provides unbuffer utility
       - python-lxml  # because pip installation is slow
+      - language-pack-br
 #      - libxml2-dev
 #      - libxmlsec1-dev
 #      - libxslt-dev
@@ -34,11 +35,7 @@ install:
   - travis_install_nightly
   - pip install -r requirements.txt
 
-before_script:
-  - chmod +x configure_locale.sh
-
 script:
-  - sudo ./configure_locale.sh
   - travis_run_tests
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,8 @@ env:
   global:
   - VERSION="10.0" TESTS="0" LINT_CHECK="0"
   - WKHTMLTOPDF_VERSION="0.12.5"
+  - LANG=pt_BR.UTF-8
+  - LC_ALL=pt_BR.UTF-8
 
   matrix:
   - LINT_CHECK="1"


rvalyi:
So I tried the language-pack-br suggestion from https://github.com/OCA/l10n-brazil/pull/733 to fix the Runbot on 10.0 but sadly this won't work. So basically it allows Runbot to keep going because no sudo is needed. However language-pack-b does not what the configure_locale.sh script was meant to do. So basically it looks pybrasil won't install (and probably pysped won't either). Indeed I even tried to larger changeset with https://github.com/OCA/l10n-brazil/pull/746 so many modules would be installed in the Runbot. But we see in the logs https://runbot.odoo-community.org/runbot/static/build/3392198-746-bc3a25/logs/job_20_test_all.txt that l10n_br_resource fails to install because of the pybrasil external dependency. The error message is a bit mysterious but it says:
```
File "/.repo_requirements/odoo/odoo/modules/registry.py", line 83, in new
odoo.modules.load_modules(registry._db, force_demo, status, update_module)
File "/.repo_requirements/odoo/odoo/modules/loading.py", line 342, in load_modules
modules.button_install()
File "/.repo_requirements/odoo/odoo/addons/base/module/module.py", line 369, in button_install
modules.state_update('to install', ['uninstalled'])
File "/.repo_requirements/odoo/odoo/addons/base/module/module.py", line 347, in state_update
self.check_external_dependencies(module.name, newstate)
File "/.repo_requirements/odoo/odoo/addons/base/module/module.py", line 320, in check_external_dependencies
raise UserError(msg % (module_name, e.args[0]))
UserError: (u'Unable to install module "l10n_br_resource" because an external dependency is not met: unsupported locale setting', '')
```

And I think this "unsupported locale setting" comes from pybrasil because language-pack-br doesn't make it happy. I think it comes from here https://github.com/odoo-brazil/pybrasil/blob/master/pybrasil/__init__.py

Can anyone confirm my analysis? So what do we do? I suggest patching pybrasil and pysped to have a degraded version that al least installs (even if it doesn't work fully) when the pt_BR locale is not available so we can have a Runbot in 10.0 too.

cc @renatonlima @mbcosta @mileo @ananiasfilho @hendrixcosta

mileo:
@rvalyi Testing force global locals

mileo:
Hi, we are working to replace pybrasil.feriado used in l10n_br_resource with https://github.com/peopledoc/workalendar 

We will compare pybrasil features and make a pull request in workcalendar, then remove-it from 10.0+

cc: @luismalta

rvalyi:
@mileo that's cool however I fear we will have the same problem with pysped...
It looks your ENV vars approach didn't work either right? We have no Runbot log but it looks it failed the same way. If you are courageous you may try the same with a larger changeset like I did in https://github.com/OCA/l10n-brazil/pull/746 I believe we then see the Runbot logs at least...

renatonlima:
Closing this PR, runbot on 10.0 still broke, but in 12.0 is working.

PR #778, opened 2020-01-14
[12.0][new] l10n br_simple

renatonlima:
Hi guys,

Today we can't install l10n_br core module to use brazilian chart of account since 9.0 when there were many changes in account, I think the module has been migrated automatically and some information does not represent the configuration of the data in Brazil, for example the creation of account.tax.tag is not correct (it is not necessary to create tags for each percentage of taxes) Also the account.tag is not correct (account.group should be created to create the accounting account hierarchy), I created the module l10n_br_simple to add in Brazilian localization OCA/l10n-brazil:

Removed creating fields in core objects account.tax.template , account.tax and its views;
Update tax group;
Update tax template;
Update account.group to implement hierarchy in accounting reports;
Update chart of account according Brazilian Federal Accounting Council Resolution No. 1.418/12


l10n_br_simple/data/account_group_data.xml
```diff
@@ -0,0 +1,223 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo noupdate="1">
+
+    <record id="account_group_1" model="account.group">
```
mileo, line 4: Can you rename all the records with a mor comprehensive name?
For example ACCOUNT_GROUP_ATIVO

mileo (changes requested):

l10n_br_simple/data/account_group_data.xml
renatonlima: @mileo,

In this case we use in XML ref the code account, there is no reason to use account name because if you want load a custom account you cant reuse this account group in reports

mileo:
image


Ainda acho que vale considerar o external id mais coerente.

mileo (approved):

rvalyi (approved):

mbcosta (approved):

l10n_br_simple/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+{
+    "name": "Brazilian - Simple Accounting",
+    "summary": "Brazilian Simple Chart of Account",
+    "category": "Localization",
+    "license": "AGPL-3",
+    "author": "Akretion, " "Odoo Community Association (OCA)",
+    "website": "http://github.com/OCA/l10n-brazil",
+    "depends": ["account"],
```
rvalyi, line 12: @renatonlima could you add "version": "12.0.1.0.0" in the manifest? the merge is failing because it is missing.

PR #818, opened 2020-04-23
Fix Document button actions

luismalta:


mileo:
Mas e as actions que estão na pasta wizard?

mileo (approved):

rvalyi (approved):

PR #937, opened 2020-08-26
[12.0] loading l10n_br_coa using hook file

renatonlima:
To setup company the right method to load COA is try_loading_for_current_company, but this method loads the COA only for self.env.user.company_id, this PR changes the way to load COA for demo company using XML file to hook post_init_hook method.


marcelsavegnago (approved):
Functional review Ok

rvalyi (approved):

mbcosta (approved):

PR #950, opened 2020-09-15
[12.0][FIX] l10n_br_nfse: is not possible select city_taxation_code on product_template form

marcelsavegnago:
I removed this widget because I was unable to select a city tax code. Honestly, I don't know if there is a better solution but I accept suggestions.

### DIFF
--- a/l10n_br_nfse/__manifest__.py
+++ b/l10n_br_nfse/__manifest__.py
@@ -26,6 +26,7 @@
 
         'views/document_view.xml',
         'views/product_template_view.xml',
+        'views/product_product_view.xml',
         'views/document_line_view.xml',
         'views/res_company_view.xml',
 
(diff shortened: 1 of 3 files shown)


luismalta:
I really think it was a mistake of mine to use this widget. Without the widget it seems to me to be the best solution.

luismalta (approved):

gabrielcardoso21:
I accessed the runbot and in the product_product form in the Invoicing app the field city_taxation_code is still not selectable. The solution would be inherit the view l10n_br_fiscal.product_form and put the field city_taxation_code after the field service_type_id. I also realized the correction you made works, but it would be best if we used this pull request to fix this other problem too.

marcelsavegnago:
@gabrielcardoso21 Perfect. I will make the change..

marcelsavegnago:
@gabrielcardoso21 done.. i appreciate if you review

gabrielcardoso21 (approved):

PR #988, opened 2020-10-21
[14.0] l10n_br_fiscal migration

mileo:


rvalyi:
superseded by https://github.com/OCA/l10n-brazil/pull/1572 which is a fresh extraction from the 12.0 branch.
Also because we implemented #1445 in 12.0 previously, it will now be reasonable to front and backport commit between the 12.0 and the 14.0 branches (cherry-pick will work).

If somebody used this branch in production, you could adapt migration scripts taken from https://github.com/OCA/l10n-brazil/tree/12.0/l10n_br_fiscal/migrations to update for the last changes.

cc @renatonlima @marcelsavegnago @mbcosta

PR #989, opened 2020-10-21
[14.0] l10n_br_coa migration

mileo:


rvalyi:
superseded by https://github.com/OCA/l10n-brazil/pull/1585
which had the CI green and was better because it was done atop of https://github.com/OCA/l10n-brazil/issues/1445 minimizing the v12 diff and git history divergence. That's is a much easier maintenance.

PR #990, opened 2020-10-21
[14.0] l10n_br_coa_generic migration

mileo:


rvalyi:
superseded by https://github.com/OCA/l10n-brazil/pull/1589
which had the CI green and was better because it was done atop of https://github.com/OCA/l10n-brazil/issues/1445 minimizing the v12 diff and git history divergence. That's is a much easier maintenance.

rvalyi:
@mileo @netosjb @renatonlima also I looked if I could cherry-pick your commits but that was not possible because of all the latest v12 changes https://github.com/OCA/l10n-brazil/commits/12.0/l10n_br_coa_generic/data. These changes show how important was the pre-commit formatting on v12, because once it's done we can front and back port most of such changes. But before the pre-commit, no way...

PR #1101, opened 2021-01-22
[12.0][l10n_br_base][OpenUpgrade] wrong bank xmlid rename; fixed no city

rvalyi:
Not sure if it's meant for merge or just to help people migrating in a particular situation:

basically I had a problem with the previous bank xmlids migration:
```
  File "/odoo/external-src/l10n-brazil/l10n_br_base/migrations/12.0.1.0.0/pre-migration.py", line 130, in migrate
    openupgrade.rename_xmlids(env.cr, _xml_ids_banks_renames)
  File "/src/openupgradelib/openupgradelib/openupgrade.py", line 964, in rename_xmlids
    logged_query(cr, query, tuple(new.split('.') + old.split('.')))
  File "/src/openupgradelib/openupgradelib/openupgrade.py", line 1339, in logged_query
    cr.execute(query, args)
  File "/odoo/src/odoo/sql_db.py", line 148, in wrapper
    return f(self, *args, **kwargs)
  File "/odoo/src/odoo/sql_db.py", line 225, in execute
    res = self._obj.execute(query, params)
psycopg2.IntegrityError: duplicate key value violates unique constraint "ir_model_data_module_name_uniq"
DETAIL:  Key (name, module)=(res_bank_104, l10n_br_base) already exists.```

So I decided to adapt the xmlid dynamically from the code_bc of the records instead.

Also, I had an issue with cities when city_id could be NULL for some international partner.
I also had an other issue with the cities migration though I now cannot remember exactly what was the problem. Will comment if I remember.

### DIFF
--- a/l10n_br_base/migrations/12.0.1.0.0/post-migration.py
+++ b/l10n_br_base/migrations/12.0.1.0.0/post-migration.py
@@ -1,30 +1,24 @@
-# -*- coding: utf-8 -*-
 # Copyright 2018 KMEE INFORMATICA LTDA - Gabriel Cardoso de Faria
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 from openupgradelib import openupgrade
 
-_model_renames = [
-    ('l10n_br_base.city', 'res.city'),
-]
-
-_table_renames = [
-    ('l10n_br_base_city', 'res_city'),
-]
-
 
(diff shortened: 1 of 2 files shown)


marcos-mendez (approved):
Por mim ta ok, testei e funfou

marcos-mendez (approved):
Por mim ta ok, foi aqui

PR #1205, opened 2021-03-13
[12.0] l10n_br_account inherits aml

mileo:
Depends of #1125


mileo:
Hi Guys, 

This concept is the sync of aml with fiscal payment of #1125. There are other ways to do that, like as done in the old #806 

Can we discuss the best way to do it? 

We propose the #820 a year ago, can you review it?

The only fresh feed back is https://github.com/OCA/l10n-brazil/pull/849#pullrequestreview-587643003 

Of course we can remove the account.payment.term copy, the user will lost some usability (only manual lines), but this can be done in a separate module.

PR #1209, opened 2021-03-15
[12.0][FIX] NFS-e document number

gabrielcardoso21:
In the NFS-e documents, both the fields `rps_number` and `number` got their values from the document_serie_id sequence, causing this document number to advance two by two. This Pull Request solves it.

### DIFF
--- a/l10n_br_nfse/models/document.py
+++ b/l10n_br_nfse/models/document.py
@@ -90,6 +90,7 @@ class Document(models.Model):
                     if not record.rps_number and record.date:
                         record.rps_number = record.document_serie_id.\
                             next_seq_number()
+                        record.number = record.rps_number
         super(Document, self - self.filtered(fiter_processador_edoc_nfse)
               ).document_number()
 
--- a/l10n_br_nfse/models/document.py
+++ b/l10n_br_nfse/models/document.py
@@ -90,7 +90,8 @@ class Document(models.Model):
                     if not record.rps_number and record.date:
                         record.rps_number = record.document_serie_id.\
                             next_seq_number()
-        super(Document, self).document_number()
+        super(Document, self - self.filtered(fiter_processador_edoc_nfse)
+              ).document_number()
 
     def _generate_key(self):
         remaining = self - self.filtered(fiter_processador_edoc_nfse)
--- a/l10n_br_nfse_ginfes/tests/test_fiscal_document_nfse_ginfes.py
+++ b/l10n_br_nfse_ginfes/tests/test_fiscal_document_nfse_ginfes.py
@@ -34,7 +34,8 @@ class TestFiscalDocumentNFSeGinfes(TestFiscalDocumentNFSeCommon):
         self.nfse_same_state._onchange_document_serie_id()
         self.nfse_same_state._onchange_fiscal_operation_id()
         self.nfse_same_state._onchange_company_id()
-        self.nfse_same_state.rps_number = 50
+        self.nfse_same_state.rps_number = '50'
+        self.nfse_same_state.number = '50'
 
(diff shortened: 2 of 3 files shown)


mileo (approved):

renatonlima (changes requested):
you have same value for rps_number and number, why do you create rps_number field? Some cities allow you to have different sequences for RPS and NFS-e, but I looked at the NFS-e modules and today this field is simply copied. IMPO if you have same value in both fields maybe rps_number field should be related of number and when you implement different sequence for RPS you can use a concrete field.

mileo:
> you have same value for rps_number and number, why do you create rps_number field

@renatonlima 

Trying to explain it better, we need both numbers:

1. RPS: Local sequence;
2. Document Number: Remote Web-service Number;

User can check both on provider web user interface. 

The point is that the fiscal document number, was being assigned by Odoo, and should be attributed by the return of the web-service, causing a double advance (One for RPS and other for NFS-e).

mileo:
@gabrielcardoso21 can you check the travis error?

marcelsavegnago:
> > you have same value for rps_number and number, why do you create rps_number field
> 
> @renatonlima
> 
> Trying to explain it better, we need both numbers:
> 
> 1. RPS: Local sequence;
> 2. Document Number: Remote Web-service Number;
> 
> User can check both on provider web user interface.
> 
> The point is that the fiscal document number, was being assigned by Odoo, and should be attributed by the return of the web-service, causing a double advance (One for RPS and other for NFS-e).

De fato o número da NFSe e do RPS é diferente.

mileo:
@gabrielcardoso21 esse processador foi deprecado no pr https://github.com/OCA/l10n-brazil/pull/1131/files

renatonlima:
@mileo,

I know the diference between RPS and NFS-e number, my question is why write rps_number in number when you confirm Fiscal document? I saw that the number is set in the return:

https://github.com/OCA/l10n-brazil/blob/bb05bff31331071bd0c93378f9c8c5c5fdd23bb1/l10n_br_nfse_ginfes/models/document.py#L341

The problem with this is receivable and the accounting transaction will have the wrong document number, maybe the RPS should be sent before and after the tax document is confirmed.

I created this issue to describe this change https://github.com/OCA/l10n-brazil/issues/1213

For now we can keep the two fields the same.

gabrielcardoso21:
> @mileo,
> 
> I know the diference between RPS and NFS-e number, my question is why write rps_number in number when you confirm Fiscal document? I saw that the number is set in the return:
> 
> https://github.com/OCA/l10n-brazil/blob/bb05bff31331071bd0c93378f9c8c5c5fdd23bb1/l10n_br_nfse_ginfes/models/document.py#L341
> 
> The problem with this is receivable and the accounting transaction will have the wrong document number, maybe the RPS should be sent before and after the tax document is confirmed.
> 
> I created this issue to describe this change #1213
> 
> For now we can keep the two fields the same.

The problem is that the field number is used in all the mixins of the fiscal document and letting it empty triggers a lot of errors. One solution to make it right once and for all is use the RPS number as the field number and store the number that comes from the response in another field. @renatonlima what do you think?

renatonlima:
@gabrielcardoso21,

It's not a good idea replace the NFS-e number coming from web service in number field in odoo because this number will used for fiscal and account reports,  now I understand the reason to copy rps_number in number, I think we can keep this implementation for now to avoid errors on actual workflow and after we can improve this as a describe in  #1213

renatonlima:
@gabrielcardoso21,

Can you check travis? 

https://travis-ci.com/github/OCA/l10n-brazil/jobs/491441874#L2611

renatonlima (approved):

PR #1224, opened 2021-03-18
[12.0][ADD] Sale operation to final consumption

gabrielcardoso21:
This Pull Request adds a separate sale operation to final consumption.

### DIFF
--- a/l10n_br_fiscal/data/operation_data.xml
+++ b/l10n_br_fiscal/data/operation_data.xml
@@ -32,6 +32,7 @@
         <field name="fiscal_operation_type">out</field>
         <field name="fiscal_type">sale</field>
         <field name="default_price_unit">sale_price</field>
+        <field name="ind_final">0</field>
         <field name="state">approved</field>
     </record>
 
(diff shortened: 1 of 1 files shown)


l10n_br_fiscal/data/operation_data.xml
```diff
@@ -83,6 +84,36 @@
         <field name="state">approved</field>
     </record>
 
+    <record id="fo_venda_final" model="l10n_br_fiscal.operation">
+        <field name="code">Venda Consumo Final</field>
```
marcelsavegnago, line 88:
```suggestion
        <field name="code">Venda Consumidor Final</field>
```

l10n_br_fiscal/data/operation_data.xml
```diff
@@ -83,6 +84,36 @@
...
+        <field name="fiscal_operation_type">out</field>
+        <field name="fiscal_type">sale</field>
+        <field name="default_price_unit">sale_price</field>
+        <field name="ind_final">1</field>
+        <field name="state">approved</field>
+    </record>
+
+    <record id="fo_venda_final_venda" model="l10n_br_fiscal.operation.line">
+        <field name="fiscal_operation_id" ref="l10n_br_fiscal.fo_venda_final"/>
+        <field name="name">Venda Consumo Final</field>
```
marcelsavegnago, line 99:
```suggestion
        <field name="name">Venda Consumidor Final</field>
```

l10n_br_fiscal/data/operation_data.xml
```diff
@@ -83,6 +84,36 @@
...
+        <field name="cfop_internal_id" ref="l10n_br_fiscal.cfop_5101"/>
+        <field name="cfop_external_id" ref="l10n_br_fiscal.cfop_6101"/>
+        <field name="cfop_export_id" ref="l10n_br_fiscal.cfop_7101"/>
+        <field name="state">approved</field>
+        <field name="product_type">04</field>
+    </record>
+
+    <record id="fo_venda_final_revenda" model="l10n_br_fiscal.operation.line">
+        <field name="fiscal_operation_id" ref="l10n_br_fiscal.fo_venda_final"/>
+        <field name="name">Revenda Consumo Final</field>
```
marcelsavegnago, line 109:
```suggestion
        <field name="name">Revenda Consumidor Final</field>
```

l10n_br_fiscal/data/operation_data.xml
```diff
@@ -83,6 +84,36 @@
         <field name="state">approved</field>
     </record>
 
+    <record id="fo_venda_final" model="l10n_br_fiscal.operation">
+        <field name="code">Venda Consumo Final</field>
+        <field name="name">Venda Consumo Final</field>
```
marcelsavegnago, line 89:
```suggestion
        <field name="name">Venda Consumidor Final</field>
```

marcelsavegnago (changes requested):

l10n_br_fiscal/data/operation_data.xml
```diff
@@ -83,6 +84,36 @@
...
+        <field name="fiscal_operation_type">out</field>
+        <field name="fiscal_type">sale</field>
+        <field name="default_price_unit">sale_price</field>
+        <field name="ind_final">1</field>
+        <field name="state">approved</field>
+    </record>
+
+    <record id="fo_venda_final_venda" model="l10n_br_fiscal.operation.line">
```
renatonlima, line 97: Why the reason for split Sale and Sale final consumption? Today we have Sale operation and many operation line for final consumption and non-consumption. This type of change only makes complex parameterization unnecessary and makes it difficult to view reports...

renatonlima (changes requested):

l10n_br_fiscal/data/operation_data.xml
gabrielcardoso21: The `ind_final` field only exists in the `l10n_br_fiscal.operation` model and not in the `l10n_br_fiscal.operation.line`. In the taxes calculation, the parameter used for making the decision of final consuption is the operation field `ind_final`. That's why we needed to add this extra duplicate sale operation to make the system able to calculate the correct taxes. I agree that it should be as simple as possible for the final user, but this would require a change in the taxes calculation. More specifically here https://github.com/OCA/l10n-brazil/blob/12.0/l10n_br_fiscal/models/tax.py#L362.

l10n_br_fiscal/data/operation_data.xml
gabrielcardoso21: I don't know if I agree with this change. The fiscal configuration of final consumption is broader than that. It can be a sale to a final consumer or to a company who will not sell the product again. In such case the product will be bought by a company, who will not consume the product, but make it available for its employees to consume it. The sale will be for final consumption, but not to the final consumer per se.

l10n_br_fiscal/data/operation_data.xml
mileo: Guys could you check what we had done on v10:

https://github.com/kmee/l10n-brazil/blob/10.0-develop/sped_imposto/models/sped_operacao.py#L210-L214
https://github.com/kmee/l10n-brazil/blob/10.0-develop/sped_imposto/models/sped_operacao_item.py#L47-L51

PR #1234, opened 2021-03-19
[12.0][ADD] Invisible fields to persist sale line taxes

gabrielcardoso21:
The values of the taxes bases and values change with the discount onchange, but are not persisted to the lines. The result is that when we invoice the sale order (directly or through stock pickings), the values in the invoice are correct, but the taxes bases and values in the invoice lines are wrong. This Pull Request adds the missing fields invisible in the sale order line tree view so they can be persisted in the database.

### DIFF
--- a/l10n_br_sale/views/sale_view.xml
+++ b/l10n_br_sale/views/sale_view.xml
@@ -57,7 +57,18 @@
                 <field name="fiscal_tax_ids" widget="many2many_tags" options="{'no_create': True}"/>
             </xpath>
             <xpath expr="//field[@name='order_line']/tree/field[@name='price_subtotal']" position="after">
-                <field name="price_total" invisible="1"/>
+                <field name="cofins_base" invisible="1"/>
+                <field name="cofins_value" invisible="1"/>
+                <field name="icms_base" invisible="1"/>
+                <field name="icms_value" invisible="1"/>
+                <field name="icmsst_base" invisible="1"/>
+                <field name="icmsst_value" invisible="1"/>
+                <field name="ipi_base" invisible="1"/>
+                <field name="ipi_value" invisible="1"/>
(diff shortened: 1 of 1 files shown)


mileo:
@gabrielcardoso21 vale fazer a mesma coisa para os outros módulos purchase / repair e etc

mileo (approved):

mileo:
Vale tb dar uma revisa aqui: 

https://github.com/OCA/l10n-brazil/blob/12.0/l10n_br_fiscal/views/document_fiscal_line_mixin_view.xml

Ver se vc não esqueceu nenhum campo.

rvalyi:
ta meio esquezito accrescentar esses campos um a um, nao deveria sera feito pelo fields_view_get do mixin?

renatonlima (changes requested):
Eu não estou certo da necessidade de existir esses campos na visão tree...

marcelsavegnago:
> ta meio esquezito accrescentar esses campos um a um, nao deveria sera feito pelo fields_view_get do mixin?

Pensei a mesma coisa.

gabrielcardoso21:
> Eu não estou certo da necessidade de existir esses campos na visão tree...

These fields exist in the form view and are not being persisted. If there's another way of persisting them without they're present in the tree view I'll be glad to change the approach.

gabrielcardoso21:
> ta meio esquezito accrescentar esses campos um a um, nao deveria sera feito pelo fields_view_get do mixin?

The problem is that there is just a few fields that need to be included. In theory, the only ones that need to be included are the base and the value of the taxes. But it sure can be done, even with a search in line mixin fields terminated in `'_base'` or `'_value'`, I'll look into it. Thank you.

gabrielcardoso21:
Solved by: https://github.com/OCA/l10n-brazil/pull/1175

PR #1270, opened 2021-04-08
[12.0][IMP] Variables for ICMSSN

ygcarvalh:
Good afternoon people.

This PR adds some variables that were missing for the calculation of the taxation of Simples Nacional and was affecting the generation of the fields of the NF-e when the CSOSN was worth 500 or 900.

I also put these fields in the view and the conditions for them to appear only when they are needed.

PR #1277, opened 2021-04-15
[IMP] Mock zip tests

mileo:
Some times travis fail because some issue with the Correios Webservice.

### DIFF
--- a/l10n_br_zip/tests/test_l10n_br_zip_res_company.py
+++ b/l10n_br_zip/tests/test_l10n_br_zip_res_company.py
@@ -2,8 +2,16 @@
 #   Magno Costa <magno.costa@akretion.com.br>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+from unittest import mock
 
 from odoo.tests.common import TransactionCase
+
+_module_ns = 'odoo.addons.l10n_br_zip'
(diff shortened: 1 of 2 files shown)


rvalyi:
I'm not so sure if it's a good idea in the case of the correios. All in all it hardly fails once every 100 Travis run or more. The mock would bring this down to zero, but it will also fail silently if there is a webserice change as it happened a few months ago. So all in all I'm not in favor of the change here. (it's not the same with the NFe, for once because not mocking would slow down a lot more and also because everybody is watching after any possible NFe webservice change so they wouldn't be missed more than a day anyway).

mileo:
Ocorreu no merge desse PR:

Link correto: https://github.com/OCA/l10n-brazil/pull/1270

https://travis-ci.com/github/OCA/l10n-brazil/jobs/499482447#L3969

rvalyi (approved):
Bom realmente tou vendo mais problemas com os site dos correios ultimamente, entao bora fazer o merge desse. Porem talvez precisa desse primeiro https://github.com/OCA/l10n-brazil/pull/1304
Vamos ver se esse vai se nao a gente faz um PR combinado. Tb teve falha com l10n_br_website_sale no ultimo run entao estamos cercados...

renatonlima (approved):

PR #1299, opened 2021-04-19
[12.0][FIX] Reports not finding field fiscal_operation_line_id

gabrielcardoso21:
While updating an old database to the current version of OCA 12.0 branch, the system couldn't find the field fiscal_operation_line_id when loading the reports of l10n_br_sale and l10n_br_purchase. I had to copy the field from the mixin to the python file in the models sale.order.line and purchase.order.line in those modules to be able to update them. If anyone knows any better way of solving this, please tell us. If not, this pull request solves the problem.

### DIFF
--- a/l10n_br_purchase/models/purchase_order_line.py
+++ b/l10n_br_purchase/models/purchase_order_line.py
@@ -31,6 +31,12 @@ class PurchaseOrderLine(models.Model):
         domain=lambda self: self._fiscal_operation_domain(),
     )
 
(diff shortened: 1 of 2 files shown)


mileo (approved):

marcelsavegnago (approved):

mileo:
@gabrielcardoso21 Could you put a comment near the field?

mileo:
> @gabrielcardoso21 Could you put a comment near the field?

UP

PR #1305, opened 2021-04-23
[12.0] [IMP] Attributes for the Simples Nacional

ygcarvalh:
Hey guys.

The PR is related to the issue https://github.com/OCA/l10n-brazil/issues/1298.

Regarding the variables that have been added the user must make the insertion manually, since it will only used when there is resale of goods with tax substitution (CST 60 or CSOSN 500).

What has been done so far is:

- addition of the percentage for Simples Nacional when it is in Range 6 of Annex 1;
- in the final value of the product, the withdrawal of credit from Simples Nacional;
- some condition for viewing the fields.

What still needs to be done is in relation to the view conditions of the fields for certain situations. It is also necessary to have a better understanding of the use of Simples Nacional credit for different tax operations.

cc: @mileo @luismalta @gabrielcardoso21 @renatonlima @marcelsavegnago


ygcarvalh:
Hey guys. During the weekend, I studied a little more about the operation of Simples Nacional and managed to draw some conclusions about the variables and the use of credit.

- In one of the commits - which I already made the correction - I stated that the credit was deducted from the value of the product, but this value has no impact on the invoice values, being used later by the acquiring company if it intends to commercialize or industrialize the merchandise. acquired.

- For the new variables that were added at https://github.com/OCA/l10n-brazil/pull/1270, they must be entered manually by the user, as it will depend on whether it is for the end consumer or not.

- For example, if there is a resale of goods with tax substitution, if the operation is not for an end consumer, the fields “ICMS ST WH Base”, “ICMS ST WH%”, “ICMS ST WH Value” and “ICMS Substitute ”Must be completed, in addition to the FCP values. However, the fields in the “ICMS Effective Calculation” group must remain unchanged, since they are used only when it is an operation for the final consumer.

- Another point was that I withdrew the DIFAL group, which started not to be charged as of 2016 with Liminar 5464 for companies opting for SN, but it continues for other types of companies. Therefore, the FCP group was created only when it has application in the context - some CSOSN codes.

ygcarvalh:
Can you take a look, please?

cc: @mileo @gabrielcardoso21 @luismalta

rvalyi:
Pessoal, eu do os puxões de orelha em inglês para nenhum 71 tentar re-escrever a historia e enrolar o pessoal na OCA e botar o nariz na merda de quem pisa na bola, mas para esse tipo de PR nao vejo vantagem nenhum em conversar sobre esse tipo de PR em ingles...

ygcarvalh:
Foi adicionado o cálculo para o valor do FCP quando há substituição tributária. Para isso, foi necessário trocar o número da sequência do _tax.group_, pois é preciso a base do ICMS ST para esse calculo e, da maneira como estava anteriormente, as taxas do FCP estavam sendo calculadas antes do ICMS ST. Essa alteração fez com que não fosse necessária recalcular o valor da base dentro da função do FCP.

Também alterei algumas condições das views - tanto para o Simples Nacional quanto para empresas Normais - já que há relação entre os códigos do CST e CSOSN.

Se ficar alguma dúvida, fico disponível para explicar as mudanças de maneira bem mais específica.

ygcarvalh:
cc: @mileo

mileo (approved):

gabrielcardoso21 (approved):

PR #1393, opened 2021-05-24
[12.0][ADD] Group pickings by fiscal operation

gabrielcardoso21:
After the discussion in the pull request #1322 I realized there's still a problem in grouping different pickings in one invoice. The original workflow doesn't consider the fiscal operation when grouping pickings, even when the grouping method chosen is `'fiscal_operation'`. This Pull Request solves that.

### DIFF
--- a/l10n_br_stock_account/wizards/stock_invoice_onshipping.py
+++ b/l10n_br_stock_account/wizards/stock_invoice_onshipping.py
@@ -130,3 +130,22 @@ class StockInvoiceOnshipping(models.TransientModel):
                 key = (key, move.fiscal_operation_line_id)
 
         return key
+
+    @api.multi
+    def _get_picking_key(self, picking):
+        """
+        Get the key for the given picking.
+        By default, it's based on the invoice partner and the picking_type_id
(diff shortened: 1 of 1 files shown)


renatonlima (changes requested):
As @bmessiaz tested in older PR https://github.com/OCA/l10n-brazil/pull/1322#issuecomment-846286309 you can create invoice using grouping Products key, because it works exactly group products by Fiscal Operation Line, in brazilian localization we check to not group products with different Fiscal Operation Line https://github.com/OCA/l10n-brazil/blob/12.0/l10n_br_stock_account/wizards/stock_invoice_onshipping.py#L123

In this PR you are including fiscal_operation_id in picking group key it is not correct because fiscal_operation_id in sale.order, purchase.order, stock.picking e etc is used only by default in create document lines, what matters are the operations of documents lines. It will block create invoice with a sale and bonus picking and today it's works including the correct accounting move entry.

rvalyi:
@gabrielcardoso21 @bmessiaz @mileo 

podemos fechar esse PR entao?

PR #1433, opened 2021-06-09
[12.0][l10n_br_base] 12.0 pre commit l10n br base

rvalyi:
As per versions 13 (Akretion won't focus on it) and mostly 14 (Akretion will focus on it), passing the OCA pre-commit hook (see https://odoo-community.org/blog/the-oca-blog-1/post/black-isort-pre-commit-97 ) will be mandatory. Our experience with Shopinvader and the one from Acsone with OCA/mis-builder teaches us that activating the pre-commit (mostly black and prettier auto-formatting) on v12 already will allow us to maintain v12 and v14 in parallel by making it easy to front port and back-port PR's 1 by 1 on both versions as much as possible, at least for several months. To be clear, once the code is formatted the same on all branches, 80% you will be able to cherry-pick a commit automatically and even when it is not automatic most of  the time it will be possible with a simple amend.

Especially we refactored the localization on v12 in such a way that the main modules of OCA/l10n-brazil will differ only by a few percents between v12 and v14 as this is the way with many other mature OCA modules now that the "new Odoo API" (the one from v8) is pretty stable.

In our vision, the OC/l10n-brazil consolidation on v12 is recent because of the ambitious refactoring with the l10n_br_fiscal extraction and the new XML databinding framework. For this reason, some companies got in production on v12 only recently, so they won't be able to migrate to v14 in the very next months, so maintaining v12 for a while will be required. Biting the pre-commit bullet now just serves this goal.

We want to do this for the modules we are leading. we understand other module author may not like the idea. Basically everybody will be free to activate the pre-commit or not for his module on v12 so we don't force anybody.

See also the same logic behind the now useless @api.multi removal https://github.com/OCA/l10n-brazil/pull/1399
and the numerous lint PR merged over the last 3 weeks to get these pre-commit passing (it also allowed us to detect some bugs early).


marcelsavegnago (approved):
Muito bom.. entendo a importancia e concordo.

rvalyi:
> Muito bom.. entendo a importancia e concordo.

Que bom que vc enxerga isso do mesmo jeito @marcelsavegnago ! Sei que hoje pode ser um pouco chato com os merges dos atuais PRs, porem quando mais cedo essa padronizaçao acontece, menos traumatismo no geral... Ate onde a gente viu vai ser bem tranquilo para manter depois.

Alem da simples questão da facilidade para manter, vc tem que ver que isso é a garantia tb que a empresa que vai migrar para a v14 daqui 2 anos nao tenha que adaptar tudo para uma localizaçao onde liberou geral para mudar tudo sem jutisficativa (aqui tem ainda os bonitoês que acham que cada nova versao é bola no centro para re-esccrever o passado, se importando bem pouco de quem apostou na OCA e financiou isso tudo).

renatonlima (approved):

renatonlima (approved):

mileo (approved):

PR #1492, opened 2021-07-01
[12.0][spec_driven_model] 12.0 pre commit spec_driven_model

rvalyi:
Same reasoning as #1433 but with the module spec_driven_model.

Additionally:

- [x] 1. unlike what you may think there is no licensing change, the module was LGPL since the release, I just fixed some wrong header
- [x] 2. clean the code so we could get rid of the useless _concrete_rec_name attributes in the spec models files like l10n_br_nfe_spec -> drier code.
- [x] 3. clean the code, specially order a bit and rename a few internal in spec_export.pý, made spec_view.py more modular.

Hint: you better check each commit to review the PR without the Black re-formatting pollution.


rvalyi:
More explanations about  the https://github.com/OCA/l10n-brazil/pull/1492/commits/1bb38feaca2639ad7b80c01d2094679b4b046839 refactor:

- clarify what is @api.model or what requires self.ensure_one()
- avoid ambiguous naming such as object -> prefer class or instance names (binding_instance is now a consistent naming accross the module for instance)
- avoid coupling the naming too much with the generateds specific databinding tool
- order methods to put class methods (api@model) 1st, recordset methods 2nd, as it is usually done in Odoo modules.

renatonlima (approved):

marcelsavegnago (approved):

PR #1423, opened 2021-06-04
[12.0] [l10n_br_fiscal] lint

rvalyi:
more l10n_br_fiscal lint

@mileo as you can see, most of the code we still had to lint heavily comes from your company KMEE and was approved and pushed by you here https://github.com/OCA/l10n-brazil/pull/928
along with lot's of other very shitty untested code (models/mdfe, models/closing.py). You ended up fixing some of it, but 1 year later attachment.py was still all shitty...

This is really a problem you abuse your seat in the project PSC to push that kind of unsollicited code in the OCA and specially if you push it inside a critical module we wrote at 90% over many years without even getting our approval!! Please read again your code of l10n_br_fiscal/models/attachments.py before I tried to do something about it (hard to do anything as it is unused and untested code...)
https://github.com/akretion/l10n-brazil/blob/dc2436358509f77c3a8e51affc7bdf98884eb936/l10n_br_fiscal/models/attachment.py

Come one, is this 2020/2021 OCA quality code?? In which other OCA repo are such "contributions" accepted?
you have tests about variables like `value` or `config_ids` or `active_id` or `active_model` that you affect 3 lines before so your tests are useless (except from bloating the complexity beyond the flake8 OCA limits) and then you do nothing with these variables...

Moreover this kind of untested feature could perfectly be added in other modules. This way people could choose if they like your implementation or... may be not! But forcing us to bloat it inside the l10n_br_fiscal module and maintain it is not nice! Sorry to tell it loud and in plain English, but sadly you have been warned several times and you play it like you don't get it...

From what I saw, attachment,py seems somewhat tested in your other PR https://github.com/OCA/l10n-brazil/pull/1355/files so may be you can test if my refactor attempt didn't break it.
I made e comment BTW because I doubt your code works https://github.com/OCA/l10n-brazil/pull/1355/files#r645291763

PR #1567, opened 2021-08-17
[12.0[l10n_br_fiscal]][REM] remove unused/untested/broken l10n_br_fiscal.attachment

rvalyi:
1 year after its addition this code is:

- unused in the whole repo.
- not even used by a PR passing tests
- untested
- broken
- ugly code

Additionally, it has been kind of "brute force" into l10_br_fiscal by @mileo in https://github.com/OCA/l10n-brazil/pull/928 while the PR was done by a KMEE employee @gabrielcardoso21 and not even revised by @renatonlima the l10n_br_fiscal module main author and maintainer over the 10 last year, nor even by me... The code was even worse and **VERY BROKEN** when it was added. I tried somewhat to clean it to pass the pre-commit and commented about that back then https://github.com/OCA/l10n-brazil/pull/1423

It's worth mentioning PR https://github.com/OCA/l10n-brazil/pull/928 added ~1000 lines of Python +XML code without a single test when it was merged by @mileo. 1 year after the code (like attachment,py) added by this PR is still untested...

Supposing such code can be fixed in the future:

- it's not essential at all, it can perfectly be added in an optional module. We made dozens implementations in for companies from the Simple and Normal regime without requiring any of this.
- l10n_br_fiscal is already a very bloated module with 9172 lines of Python alone, Some features like this SHOULD be done in optional modules instead. Saying it's orthogonal to several fiscal documents is not a valid justification. A well designed module can serve in several modules without being in their dependency path. A good example of that is the OCA base_tier_validation_formula which can be used in sale_tier_validation or purchase_tier_validation or other tier* modules without  being a hard dependency for them.
- a better implementation can probably be done depending on the dms module. But again, in an optional module.

Instead there are a few small features that are completely interdependent with l10n_br_fiscal and cannot reasonably go anywhere else. So I suggest we move out what is useless and what can go in other modules to keep the codebase clean and welcoming instead of bloated. l10n_br_fiscal is already the largest OCA module by a fait margin, New features should be added with a lot of care, certainly not overnight without the main module authors approval.


rvalyi:
when I say files added in #928 are still largely broken one year later. One can see that all buttons/actions in these views 

- https://github.com/OCA/l10n-brazil/blob/12.0/l10n_br_fiscal/views/dfe/dfe_views.xml
- https://github.com/OCA/l10n-brazil/blob/12.0/l10n_br_fiscal/views/dfe/dfe_views.xml

are not implemented at all. Yes there is a broken PR for them that would require a ton of fixes before eventually making it. But I think the OCA is not about fireworks coding, one cannot invade someone else module and bloat it with broken and untested features just to fake they kind of co-authored the module. Sadly a reccuring attitude from this company and the recent attempt to re-implement the anglo-saxon accounting was just a very extreme example of such behavior, but it was certainly not the first time, far from it. Quality matters in the OCA, we cannot accept people use the project just to promote their brand with empty or useless features, it harms the maintainance, the migration and make the project look ridiculous to serious people looking at the code.

But OK, in any case I'm not even proposing to remove dfe.py and mdfe,py, just attachment,py.

renatonlima (approved):

mbcosta (approved):
ola @rvalyi 
Realmente se um código não é usado não tem sentido estar no repo, pelo que vi existe apenas um PR em Rascunho/Draft onde isso está sendo usando https://github.com/OCA/l10n-brazil/pull/1355/files#diff-672e32bacff795085ae1a2e7230f6c8403c19268c09be1d21c67ce7a2ab4ef8fR114 , e a funcionalidade de "Compactar os anexos recebidos e os retorno como um novo único anexo" pode, e é melhor, ser feita em um modulo especifico para isso, até para poder ser reutilizada se necessário por outros módulos sem necessidade de instalar o l10n_br_fiscal.

marcelsavegnago (approved):

PR #1586, opened 2021-08-23
[12.0] nfelib 1.1 pypi

rvalyi:
Same as https://github.com/OCA/l10n-brazil/pull/1575 but with nfelib 1.1 (including tests NFe XMLs in the wheel package) and new PR because #1575 was stalled on Travis.

nfelib is pretty solid right now so I cut a 1.1 release for pypi. This will help the l10n_br_nfe migration where the python package come from the module manifests and help noob installations.


rvalyi:
looks like #1575 was stalled because of my .travis.yml fix attempt

PR #1801, opened 2022-02-11
[IMP] l10n_br_fiscal: Hide page if company is not in Brazil

max3903:


felipemotter (approved):

antoniospneto (approved):

rvalyi (changes requested):
Hello @max3903 thanks for contributing. For the feature to be working as expected we should also change the attrs of the icms_regulation_id field (in the same XML file) to make it not required if the company is not Brazil. Because currently, if you create a company outside from Brazil, even if the fiscal tab is hidden, by default this field is still required while being invisible. See this print (were I left the company in Brazil to let the tab visible):

renatonlima (changes requested):
Hello @max3903,

Can you change the actual required fields inside page Fiscal to false?

felipemotter:
@renatonlima wouldn't It be better set required = false If It isn't a brazilian Company?

renatonlima:
> wouldn't It be better set required = false If It isn't a brazilian Company?

That's exactly what I meant in last comment, to use the same domain in attrs to hide Fiscal page to implements required attrs in these required fields.

max3903:
@sandipvyas-osi Can you make the suggested changes?

max3903:
@renatonlima @rvalyi It should be good now.

marcelsavegnago (approved):
LGTM

antoniospneto (approved):

l10n_br_fiscal/views/product_product_view.xml
```diff
@@ -85,7 +94,11 @@
                             </field>
                         </group>
                         <group>
-                            <field name="fiscal_genre_id" required="1" />
+                            <field
+                                name="fiscal_genre_id"
+                                required="1"
+                                groups="l10n_br_fiscal.group_manager"
+                            />
```
rvalyi, line 101: @renatonlima @marcelsavegnago I think this group is not quite right. How could we fix it?

max3903: @rvalyi Using `attrs`...

l10n_br_fiscal/models/document_fiscal_mixin_methods.py
```diff
@@ -53,6 +53,10 @@ def _compute_amount(self):
                             values[field] += line[field.replace("amount_", "")]
             doc.update(values)
 
+    def _compute_country(self):
+        for record in self:
+            record.country_id = self.env.company.country_id.id
+
```
antoniospneto, line 69: I don't think it's a good option to take the currently active company, the company that owns the document could be another.

PR #1969, opened 2022-06-08
[14.0][IMP] l10n_br_sale: Hide fields if company is not in Brazil

max3903:
Depends on #1801

### DIFF
--- a/l10n_br_sale/models/sale_order.py
+++ b/l10n_br_sale/models/sale_order.py
@@ -300,6 +300,9 @@ class SaleOrder(models.Model):
 
         inv_ids = super()._create_invoices(grouped=grouped, final=final)
 
+        if self.country_id.id != self.env.ref("base.br").id:
+            return inv_ids
+
         # In brazilian localization we need to overwrite this method
         # because when there are a sale order line with different Document
         # Fiscal Type the method should be create invoices separated.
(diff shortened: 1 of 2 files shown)


rvalyi:
Thank you @max3903 the little issue I see with #1801 so var https://github.com/OCA/l10n-brazil/pull/1801/files#r893701610

renatonlima:
@max3903,

Can you check travis status? https://app.travis-ci.com/github/OCA/l10n-brazil/jobs/573062565#L2913

renatonlima:
@max3903,

I think you forgot define country_id field in object sale.order, You can defini country_id field in https://github.com/OCA/l10n-brazil/blob/14.0/l10n_br_fiscal/models/document_fiscal_mixin.py because this object is inherited by sale.order, purchase.order, account.move and we can reuse country_id field in other brazilian localization modules.

rvalyi:
I also mention 2 related PR/issues that we should consider having working with this current PR:

- the idea to have a group/option to avoid having to fill the product fiscal shit, even for a Brazilian user: https://github.com/OCA/l10n-brazil/pull/1592
- getting the whole fiscal dance to work with product variants properly https://github.com/OCA/l10n-brazil/issues/1892

PR #1972, opened 2022-06-09
[IMP] l10n_br_base: UI changes if partner is not in Brazil

max3903:


l10n_br_base/views/res_partner_view.xml
```diff
@@ -97,7 +101,11 @@
                 </group>
             </xpath>
             <page position="after" name="sales_purchases">
-                <page string="Fiscal" name="fiscal">
+                <page
+                    string="Fiscal"
+                    name="fiscal"
+                    attrs="{'invisible': [('country_id', '!=', %(base.br)d)]}"
+                >
```
rvalyi, line 108: @renatonlima @marcelsavegnago would such thing work for a foreign customer or supplier dealing with Brazilian company? Will we be able to set such a partner properly with no fiscal tab?

rvalyi:
More specifically, if we hide this fiscal tab if the partner is not in Brazil, may be we should have some kind of default or onchange to set the partner fiscal profile and ICMS contribution field to something like "isento" or "nao contribuinte" (I'm not sure which one it should be). I'm not sure about that, but I mean we should have fiscal documents rights when we use such foreign customer or supplier.

Screenshot-55

l10n_br_base/views/res_partner_view.xml
rvalyi: @renatonlima @marcelsavegnago @max3903 @JorgeJuarezCasai

I checked this again an I thing this is OK to hide the fiscal tab IF we set an _inverse method on the country_id so that when set outside from Brazil it sets the fiscal_profile_id to "Não contribuente" (if we don't do that, in Brazil you won't be able to properly set up a foreign customer or supplier).
What do you think?

rvalyi (changes requested):

JorgeJuarezCasai:
@rvalyi i will try to define the inverse method. As i understand "Não contribuente" is a good default value for foreigns?

rvalyi:
Yes. Of course it only matters when you will have invoices with these foreign parties from Brazil, but this is definitely a use case we should consider before merging anything like that. @renatonlima  if you can confirm...

l10n_br_base/views/res_partner_view.xml
renatonlima: It's not correct hide Fiscal tab in foreign customers because here in brazil if you do a foreign sale you need see fiscal tab. The correct is hide Fiscal tab if company's country is not Brazil.

renatonlima (changes requested):

PR #2013, opened 2022-07-12
[14.0][WIP][FIX] l10n_br_fiscal - access issues

JorgeJuarezCasai:
image

Access denied error when non br fiscal user tries to read partner data.

### DIFF
--- a/l10n_br_fiscal/security/ir.model.access.csv
+++ b/l10n_br_fiscal/security/ir.model.access.csv
@@ -69,6 +69,7 @@
 "l10n_br_fiscal_operation_document_type_user","Fiscal Operation Document Type for User","model_l10n_br_fiscal_operation_document_type","l10n_br_fiscal.group_user",1,0,0,0
 "l10n_br_fiscal_operation_document_type_manager","Fiscal Operation Document Type for Manager","model_l10n_br_fiscal_operation_document_type","l10n_br_fiscal.group_manager",1,1,1,1
 "l10n_br_fiscal_partner_profile_user","Fiscal Partner Profile for User","model_l10n_br_fiscal_partner_profile","l10n_br_fiscal.group_user",1,0,0,0
+"l10n_br_fiscal_partner_profile_any","Fiscal Partner Profile for Any User Readonly","model_l10n_br_fiscal_partner_profile",,1,0,0,0
 "l10n_br_fiscal_partner_profile_manager","Fiscal Partner Profile for Manager","model_l10n_br_fiscal_partner_profile","l10n_br_fiscal.group_manager",1,1,1,1
 "l10n_br_fiscal_document_user","Fiscal Document for User","model_l10n_br_fiscal_document","l10n_br_fiscal.group_user",1,0,0,0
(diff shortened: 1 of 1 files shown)


rvalyi:
Hello @JorgeJuarezCasai letting everybody read `fiscal.partner.profile` as you are proposing seems no big deal. As for letting everybody read `l10n_br_fiscal.document` it doesn't seem reasonable to me as anybody would suddenly be able to query the equivalent of all invoices using jsonRPC requests in the browser for instance.

Could you please elaborate where the `l10n_br_fiscal.document` access matters for you? Isn't it only when the `l10n_br_account` module is installed? (because of the _inherits I guess?) if yes, what about giving access only to the `account.group_account_readonly` group instead?

JorgeJuarezCasai:
> Hello @JorgeJuarezCasai letting everybody read `fiscal.partner.profile` as you are proposing seems no big deal. As for letting everybody read `l10n_br_fiscal.document` it doesn't seem reasonable to me as anybody would suddenly be able to query the equivalent of all invoices using jsonRPC requests in the browser for instance.
> 
> Could you please elaborate where the `l10n_br_fiscal.document` access matters for you? Isn't it only when the `l10n_br_account` module is installed? (because of the _inherits I guess?) if yes, what about giving access only to the `account.group_account_readonly` group instead?

Thank you @rvalyi i was trying options to fix the issue. I was thinking on create another group and assign them to the uses need to have access.

Main issue i have is on sales orders, i have 2 companies 1 for br and another for mx, but when someone in mx tryes to read or create a partner gets an access error y they doesn't have the BR user profiles, i don't want to assign this groups to mx users.

For now i just continue testing options.

rvalyi:
OK, in any case I'll also point to a related PR that we didn't complete https://github.com/OCA/l10n-brazil/pull/1592
We wanted to make it easy, even for a Brazilian purchase or stock user to be able create new products without having to take care about all the fiscal fields and only let them complete the product fiscal fields once the product is purchased from suppliers or even let other users do it. Eventually it is related to the other things you are trying to relax in the l10n_br_fiscal module. At least when merging something we should ensure we take care of my other PR (I'm open to suggestions of course).

rvalyi:
@JorgeJuarezCasai BTW in case you missed it I also gave feedback about another fiscal relax PR you are interested in [OCA/l10n-brazil#1972](https://github.com/OCA/l10n-brazil/pull/1972#discussion_r919442695)

renatonlima:
@JorgeJuarezCasai,

We can keep  fiscal.partner.profile but to fiscal.document the correct is create permissions in l10n_br_account to grant this permission to Account Invoice User. I did one PR to fix it #2016, can you test it?

JorgeJuarezCasai:
> @JorgeJuarezCasai,
> 
> We can keep fiscal.partner.profile but to fiscal.document the correct is create permissions in l10n_br_account to grant this permission to Account Invoice User. I did one PR to fix it #2016, can you test it?

I made test and #2016 fix my problem. Thank you @renatonlima

JorgeJuarezCasai:
closed by https://github.com/OCA/l10n-brazil/pull/2016

PR #2019, opened 2022-07-13
[14.0][FIX] l10n_br_base - related var field to cnpj

JorgeJuarezCasai:
I have issues with non brazilian parners cause the vat field was overwritten and related to the cnpj field.

### DIFF
--- a/l10n_br_base/models/res_partner.py
+++ b/l10n_br_base/models/res_partner.py
@@ -21,8 +21,6 @@ class Partner(models.Model):
     _name = "res.partner"
     _inherit = [_name, "l10n_br_base.party.mixin"]
 
-    vat = fields.Char()
-
     is_accountant = fields.Boolean(string="Is accountant?")
 
     crc_code = fields.Char(string="CRC Code", size=18)
--- a/l10n_br_base/models/res_partner.py
+++ b/l10n_br_base/models/res_partner.py
@@ -21,7 +21,7 @@ class Partner(models.Model):
     _name = "res.partner"
     _inherit = [_name, "l10n_br_base.party.mixin"]
 
-    vat = fields.Char(related="cnpj_cpf")
+    vat = fields.Char()
 
     is_accountant = fields.Boolean(string="Is accountant?")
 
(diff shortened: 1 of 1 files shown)


marcos-mendez:
@renatonlima e @rvalyi quando tive este problema (no modulo de rh foi mais ou menos isso que eu fiz mesmo) por isso aprovei. Se precisarem de ajuda em testes na base podem me marcar.

marcos-mendez (approved):
ok

l10n_br_base/models/res_partner.py
```diff
@@ -21,7 +21,7 @@ class Partner(models.Model):
     _name = "res.partner"
     _inherit = [_name, "l10n_br_base.party.mixin"]
 
-    vat = fields.Char(related="cnpj_cpf")
+    vat = fields.Char()
```
rvalyi: Hello @JorgeJuarezCasai overall fixing these little glitches here and here for a multi-countries usage was expected despite I took a great care the overall modules design to make it possible.

Your fix looks good overall, though in this case you should simply remove that vat field declaration (line 24) as no vat field override is required anymore. (the default from the base module is enough).

rvalyi (changes requested):
cc @renatonlima

JorgeJuarezCasai:
@rvalyi i was removed the field override. Thank you for your review!

l10n_br_base/models/res_partner.py
```diff
@@ -35,6 +33,25 @@ class Partner(models.Model):
 
     union_entity_code = fields.Char(string="Union Entity code")
 
+    cnpj_cpf = fields.Char(
```
renatonlima, line 36: @JorgeJuarezCasai 

I tested this PR changes on runbot, in res.partner it's fine, but if you change the vat in a res.company, it is not updated in company's related partner.

To solve this you can move this changes to https://github.com/OCA/l10n-brazil/blob/14.0/l10n_br_base/models/party_mixin.py#L17 because party.mixin is injected in res.partner and res.company

renatonlima (changes requested):

l10n_br_base/models/res_partner.py
```diff
@@ -35,6 +33,25 @@ class Partner(models.Model):
...
+    def _compute_cpf_vat(self):
+        for record in self:
+            if record.country_id.id == self.env.ref("base.br").id:
+                if record.vat and not record.cnpj_cpf:
+                    record.cnpj_cpf = record.vat
+
+    def _inverse_cpf_vat(self):
+        for record in self:
+            if record.cnpj_cpf:
+                record.vat = record.cnpj_cpf
```
marcelsavegnago, line 53: Here's a detail. When the individual partner has a parent_id, the partner.vat is the same as partner.parent_id.vat.. as far as I remember this behavior is native to Odoo. I'm not sure but probably this method can end up affecting this behavior.

PR #2758, opened 2023-10-27
[16.0] dotfiles update needs manual intervention

etobella:
Dear maintainer,

After updating the dotfiles, `pre-commit run -a`
fails in a manner that cannot be resolved automatically.

Can you please have a look, fix and merge?

Thanks,


rvalyi:
This PR is replaced by https://github.com/OCA/l10n-brazil/pull/2802 which also include all the required forward ports to pass the pre-commit and the tests.

PR #2787, opened 2023-11-11
[16.0] trivial fix, resolve #2784

### LINKED ISSUE #2784: [14.0][15.0][16.0] git checkout on windows "" error: invalid path 'l10n_br_base/migrations/14.0.2.0.0/post-migration.py ' ""
## Module
l10n_br_base

## Describe the bug
When I git checkout on windows (for Docker environment) "" error: invalid path 'l10n_br_base/migrations/14.0.2.0.0/post-migration.py ' "" because file has got a space after ".py"
https://github.com/OCA/l10n-brazil/blob/15.0/l10n_br_base/migrations/14.0.2.0.0/post-migration.py%20

## To Reproduce
**Affected versions**: 14.0 15.0 16.0

1. On windows
2. Git clone this repo
3. git checkout 14.0

rpinset:
trivial fix, resolve #2784


rvalyi:
Thanks for the fix. However here it's another issues, as per the OCA guidelines, old migration scripts should be removed when migrating. Could you do a new PR removing s the migration folder entirely please?

rvalyi (changes requested):
see my previous comment

rpinset:
Hello @rvalyi 
I could delete the old folder with the migration script, but if it is a migration in 15.0 and 16.0 would there be no other choice to make which would lead to the creation of a new folder?
It is not me who initiated these files and I have not yet looked into the version upgrade (I will observe 17.0)

rvalyi:
yes in the case we would need a new v16 migration we would create the folder again. As for now you can make a PR to remove it. Also bare in mind git only deal with files and not folder, so git wouldn't even keep an empty folder in the repo unless you put some hidden file like .gitkeep inside it. So all in all, let's just remove this folder please.

l10n_br_base/migrations/14.0.2.0.0/post-migration.py
rpinset, line 1: #2787

rpinset:
it's done @rvalyi 
https://github.com/OCA/l10n-brazil/pull/2787/commits/b49ef0de627f32785d4f6f950691dc1f31fc5de8

rvalyi (approved):
thank you @rpinset !

antoniospneto (approved):

rvalyi:
trivial, fast tracking.

/ocabot merge patch

felipemotter (approved):

rvalyi:
see my comment here https://github.com/OCA/l10n-brazil/pull/2788#issuecomment-1824700718
after fixing v15 I can help fix it for v16 too.

rvalyi:
just like for the 15.0 serie, we will first need to fix the pre-commit requirements that got increased since the last 16.0 merges and I did the required commit forward ports here https://github.com/OCA/l10n-brazil/pull/2802

rvalyi:
@rpinset we fixed the pre-commit issues in 16.0 by porting a set of missing commits from the 14.0, but now unfortunately it seems your PR here cannot be merged because the automatic rebase is failing. could you rebase atop of the updated16.0 please?


on your own 16.0 branch you could do something like:

```
git fetch origin 16.0
git rebase origin/16.0
git push -f <your_remote> 16.0
```

rvalyi:
in fact it has been fixed on v16 already thanks to the port from v14 https://github.com/OCA/l10n-brazil/tree/16.0/l10n_br_base/migrations/14.0.2.0.0 so nothing more is required: it is fixed now. Thank you for the fix anyway.

PR #2952, opened 2024-03-10
[14.0][l10n_br_sale][WORK IN PROGRESS] sale inline line edition

rvalyi:
[EDIT] 

To have a future proof design, it's important to read:

From https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0:

> Any view with groups_id on it now has to move such groups to the elements of the view. But now you can put groups attribute in a view field instead of isolating it on a view without fear of an access error. Check it in https://github.com/odoo/odoo/pull/98551.
> On the same mood, you should check if the fields protected by a group interact in some way with business logic or UI. For example, a field being part of a domain of another field, or being used in copy/defaults. An special case is the field company_id, which is always used internally for domains on many2one fields referencing multi-company aware records. On these cases, you have to add the same field twice, one with the group, and another invisible without it. Check core examples in https://github.com/odoo/odoo/pull/95729.

And so the explanation here are important too: https://github.com/odoo/odoo/pull/98551

### DIFF
--- a/l10n_br_fiscal/models/document_fiscal_line_mixin_methods.py
+++ b/l10n_br_fiscal/models/document_fiscal_line_mixin_methods.py
@@ -81,7 +81,8 @@ class FiscalDocumentLineMixinMethods(models.AbstractModel):
                 # these will only collect (invisible) fields for onchanges:
                 (
                     ".//control[@name='fiscal_fields']...",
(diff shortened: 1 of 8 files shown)


rvalyi:
replaced by https://github.com/OCA/l10n-brazil/pull/2992 which is future proof.

PR #2982, opened 2024-03-27
[14.0][l10n_br_account] account move form clean up inherited fields

rvalyi:
while working on the v16 l10n_br_account migration I spotted this:

domain and groups where added to analytic fields in the account.move.line form view. It came from v12 as it. May be it was required on v12, but on v14 it seems useless as these domain and groups are already present in the parent view: account.view_move_form in
l10n_br_account/views/account_invoice_view.xml
The form view for Brazil is already very complex, it's important to keep only what is strictly necessary.

link: https://github.com/odoo/odoo/blob/14.0/addons/account/views/account_move_views.xml#L763

As I did a POC for the sale view https://github.com/OCA/l10n-brazil/pull/2952 we would like to allow the "native" inline edition mode for account.move too as it would make the localization less invasive. A very first step is to clean up these migration left-overs. The migration of l10n_br_account to v14 was very hard and I guess we ended up not paying attention to these details....

PR #2983, opened 2024-03-27
[14.0][l10n_br_account][REF] clean up inherited fields

rvalyi:
Simplified and correct version of https://github.com/OCA/l10n-brazil/pull/2982

Some fields are defined in a superficial way in the inherited move form that is considered specific for mobile edition by Odoo but not for us in the Brazilian localization. To make it work properly we had to define some extra attributes just like in the editable tree view.

However, in v14 some of these attributes where added in the inherited form view so they can be removed from the view extension as we do here.


antoniospneto (approved):
LGTM

marcelsavegnago (approved):

mileo (approved):

PR #2986, opened 2024-03-27
[15.0][l10n_br_account][REF] clean up inherited fields

rvalyi:
forward port of #2983 (the original inherited view is the same as in 14.0)

Some fields are defined in a superficial way in the inherited move form that is considered specific for mobile edition by Odoo but not for us in the Brazilian localization. To make it work properly we had to define some extra attributes just like in the editable tree view.

However, in v14 some of these attributes where added in the inherited form view so they can be removed from the view extension as we do here.


mileo (approved):

rvalyi:
merging as it's a dump forward port

/ocabot merge minor

PR #3427, opened 2024-10-08
[14.0][spec_driven_model][l10n_br_nfe] stacking fixes

rvalyi:
- a StackedModel class has always cls._stacked True, so no need to test it
- if we don't redefine _stacking_points = {} in each stacked class, then the _stacking_points property from the common spec.mixin is used and stacking_points of unrelated classes end up in the same dict, potentially creating conflicts. This can be seen in the test_nfe_structure.py tests where the l10n_br_fiscal.document had the l10n_br_fiscal.document.related object stacking points inside its own stacking points and I had to remove them (back then, I got mad why that would be the case, I'm finally happy to understand why it happens)...
 
We might end up dropping the _stacking_points property and scope it with the schema name+version with #3424 but until then, it's better to fix it...


marcelsavegnago (approved):

rvalyi:
EDIT: I forced push with 2 important changes:

- to avoid further issues with models sharing values in _stacking_points, I made it the generic _stacking_points a frozendict: https://github.com/OCA/l10n-brazil/pull/3427/files#diff-2293fd970a03ee7c5b00a0770436899be296a3438e71e78bff1e519974d70f46R26   This is a temporary defensive action, the whole thing will hopefully be fixed in #3424 which is a more extensive piece of work...
- I removed a bunch of attributes from the 4 NFe StackedModel classes as these attributes are now properly inherited from the spec.nfe.mixin (the inheritance model was fixed a few months ago in v14: https://github.com/OCA/l10n-brazil/pull/3215 )

@marcelsavegnago could you review again please? I think this is important for your MDFe and CTe work...

antoniospneto (approved):

marcelsavegnago (approved):

marcelsavegnago:
Bora mesclar.. fiz o mesmo ajuste com o mdfe e passou nos testes (ambiente somente com o mdfe e sem nfe ainda).

marcelsavegnago:
> Bora mesclar.. fiz o mesmo ajuste com o mdfe e passou nos testes (ambiente somente com o mdfe e sem nfe ainda).

testando juntos não funciona mas o escopo da PR não era tratar o multi schema.

rvalyi:
trivial, fast tracking

/ocabot merge patch

PR #2781, opened 2023-11-09
[14.0][l10n_br_fiscal][l10n_br_nfe][l10n_br_account][l10n_br_account_nfe] importação dos account.move com as NFe's - contra-proposta

rvalyi:
As announced, here is the Akretion alternative proposal for #2763, that is for importing account.move accounting and financial data along with NFe's or fiscal documents in general.

@mileo @hirokibastos as I warned you the code is totally different from what you proposed and as you can see I could not reuse even a single line of code. It took me 4 days and much more if considering I already spent lot's of time thinking about this ahead during these years when I designed the localization and specially the l10n_br_account module architecture and data-binding framework/xsdata/nfelib. So since you announced @hirokibastos would give it a try I was absolutely convinced it was out of reach from a beginner to come any close to such design. There was no way I could spend half an hour giving advice (unlike what I did for the dfe, cte, mdfe...) here and here and converging with something as complete and consistent as I'm proposing here:

**key features:**

- [x] import NFe's along with their financial account.move (until now only l10n_br_fiscal.document was imported)
- [x] import a set of **several attachments** right from the Vendor Bills **native upload button** (unlike #2763)
- [x] import nearly all Brazilian tax values (l10n_br_nfe/models/document_line,py refactor de0d912da23c31d7b51226d72d6004b26cd0dd77 ). The tax computation is disabled so these values (amount, base, rate...) are kept as imported. This is done by extracting the bare minimum from #1412 (it's better to leave the rest up for debate in another PR, specially considering also https://github.com/OCA/l10n-brazil/pull/2559 ) and complemented with 1f07ec06d29b0719e04fc14e33ef9bfb5f3e85a0 This refactor to import all tax values is one of the reasons the diff is larger but IMHO we had to do it. The code is also a lot cleaner now and I could finally clean a useless/non orthogonal importation method introduced by Gabriel Cardoso on a hurry 3 years ago. We could merge that in a separate PR if that helps...
- [x] **accounting entries for Brazilian taxes** are properly generated (not at all the case with #2763). See next point.
- [x] account.move(.lines) are generated with odoo.test.common.Form so that **all onchanges are called as expected just like if the account.move was filled manually**. This avoids spaghetti code and glue modules that would otherwise be needed to play all the onchange of any combo of installed modules. This approach is already used in the OCA for years in modules such as OCA/rma. #2763 does not address this and open the door to lot's of logic duplication.
[...]

PR #3460, opened 2024-10-20
[14.0] nfe account move import - rebased - WIP

rvalyi:
teste de rebase de #2781; não fazer o merge.


rvalyi:
now that it is green, pushing to #2781 again...

PR #3560, opened 2024-12-18
[16.0][WIP] ref account fiscal amounts icms ded

rvalyi:
WIP don't merge


rvalyi:
not yet possible to enable these tests again

PR #3563, opened 2024-12-28
[16.0][REF][l10n_br_base][l10n_br_fiscal][l10n_br_account] tests don't depend on demo

rvalyi:
both Odoo SA [1] and the OCA [2][3] recommend that tests don't depend on demo data. Here I'm starting this cleanup on the most central modules. Benefits include:

- the test context is simpler and easier to figure out
- this avoids the demo data incompatibility hell where arbitrary modules alter demo data and depending on what modules are installed exactly your tests may fail with no clear explanation
- this allows to run fiscal tests on copy of production databases

We have long wished to achieve that in the project. However, in l10n-brazil we need to maintain both complex fiscal demo data and fiscal test fixtures. That is why until now the test fiscal transactions where based on demo data.

In this PR, I created a simple framework (with l10n_br_base/tests/tools.py and l10n_br_fiscal/tests/tools.py) where we can load the SAME fiscal demo data both for the demo mode and for the tests even on databases created without demo data (like with the --without-demo=all param). I had to clean up a bit useless relations to demo data in the fiscal demo data and work around the demo data of the product module.

As we tend to reload the same fiscal demo fixtures in several tests, tests are a bit slower. But:

- in the future they could actually be faster if we create test databases with --without-demo=all
- it's v16 which is 3x faster than v14, so not such a big deal (only a few dozens of extra seconds)
- even if tests do a bit more queries, they are much easier and faster to debug this way...

Finally, let me remind you that we cannot simply use the production set up for the Brazilian tax engine both for demo and tests because unlike for European countries, the Brazilian tax engine setup requires loading several MB of XML and csv files which easily take 3 minutes of heavy CPU usage and that would be a terrible developper experience to eat it for any quick test or CI push. (At the moment the decision to load production data or not happens in l10n_br_fiscal/hooks.py)

- [1] https://www.odoo.sh/faq#without_demo
- [2] https://odoo-community.org/groups/contributors-15/contributors-1971837?mode=thread&date_begin=&date_end=
- [3] https://github.com/OCA/bank-payment/pull/1391

NOTE1: also see work in progress for l10n_br_sale module (some tests still failing) https://github.com/akretion/l10n-brazil/commit/9667018fe4c0e919f2e38d7ab63b1f581469581c (in fact it seems my local test with --without-demo=all failed because of discrepancies between the demo and prod NCM data. See next note.

NOTE2: this PR will also work best in conjunction with the refactor of #3567 to standardize the data and demo files loading. Once #3567 is merged we will get rid of the demo NCM and CEST files so the load of the test fixtures will be much faster!


rvalyi:
pretendo fazer um rebase e finalizar assim que https://github.com/OCA/l10n-brazil/pull/3567 entrar...

cc @mbcosta @antoniospneto

PR #3567, opened 2024-12-30
[16.0][l10n_br_fiscal][IMP] l10n_br_fiscal: standardize data+demo load

rvalyi:
Standardize the loading of all l10n_br_fiscal data + demo files.

In l10n_br_fiscal we have some very specific challenges:

- loading all the production fiscal data for the tax engine takes 3 minutes and a lot of CPU, this is a no go for demo and tests. This is why we had these subset demo files for NCM, NBM, NBS, CEST...
- for mission critical cases, some fiscal data might better be altered by some fiscal power user and keep noupdate=True. If noupdate were False, reloading the l10n_br_fiscal module would also take a lot of time. By default modules data files load with noupdate=False.

This explains the peculiar loading hook we used in l10n_br_fiscal/hooks.py until now.

But after working on the isolation of the tests from the generic demo data, I also got convinced we could finally standardize the loading of the data + demo files.

In fact all it takes is a custom monkey patch of the odoo.tools.convert.convert_csv_file method to force noupdate=True for some files and to be able to filter and load only some demo records in demo/test mode for some large csv files.

Benefits:

- standard loading of data + demo files; simpler
- pylint-odoo will now hunt for bugs in some 25 CSV and XML demo and data files that were not analysed before when they were not declared in the manifest.
- no more "file-not-used" pylint warnings we could not even filter (in v18 warnings will we block the CI by default)
- removed 800 lines of demo files we had to maintain
- no more production/demo data discrepancies
- will also help simplify the loading of fiscal fixtures in #3563

NOTES: in 2022 the TIPI data was updated but the demo TIPI csv file kept using the old IPI from 2021 or before. So now that the tests and demo are using the same production ncm CSV file (a subset of it actually) I had to update a few tests to change the IPI from 5% to 3.25% for some products.


rvalyi:
pronto para revisão. Tirou tb mais de 500 linhas, coisa que é bom no modulo l10n_br_fiscal que já ta pesado demais...

antoniospneto (approved):
Muito bom, parabéns pelo trabalho @rvalyi 
Fica bem melhor essa nova abordagem.

rvalyi:
@antoniospneto eu apenas fiz um force push para limpar o manifest https://github.com/OCA/l10n-brazil/compare/8b6098476cc28c11e350a0626c094d55fc4133c8..910d98e123fdc0a1e9e8c6afb6a8988f9f93c1b4

marcelsavegnago (approved):

rvalyi:
vamos ver se os testes vão passar...

rvalyi:
aquele problema das datas no mes de fevereiro, vamos ver amanhã...

PR #3777, opened 2025-05-07
[16.0][ADD] l10n_br_fiscal: editable data mixin

rvalyi:
Rationale:

As for today many very large fiscal data are loaded via CSV or XML. Examples:

- ncm
- cnae
- cfop
- cest
- nbs
- tax.pis.cofins
- fiscal.tax
- cst
- tax.definition

Because these files can be so huge, we often load them with noupdate=True (we even monkey_patched convert_csv_import for this) to avoid long minutes of updates in production.

Also historically these files were poorly maintained. So when Akretion had a big customer, we would maintain their data for their specific operations, but don't really bother maintaining these files up to date. This is now changing with more serious projects in production and more serious contributors participating.

A problem with these data is when a user wants to update some value manually or create a new record (like a new NCM that was missing). These records will have no xml_id or ir.model.data entry. And later when we try to update or migrate the database, the new records that would eventually be inserted in the source csv or XML files would clash with these manually entered records in the database making the whole upgrade process cumbersome.

So I introduced a new mixin here that will fix this by providing these features:

- Automatically creates ir.model.data entries for manually created records.
- Provides `get_records_without_xmlid()` to find records lacking an XML ID.
- Provides `fill_missing_xml_ids()` to backfill missing XML IDs for existing records.
- Prevent writing values inconsistent with record xml_id.
- Allow to toggle update/noupdate mode.

NOTE: some of the mixin code has been proposed by the Gemini AI (I'm still adjusting it) and most of the tests were written by the AI. This worked because this code is quite independent from the localization code.

### DIFF
--- a/l10n_br_fiscal/__init__.py
+++ b/l10n_br_fiscal/__init__.py
@@ -3,6 +3,7 @@
 
 from . import models
 from . import wizards
+from .hooks import post_init_hook
 
 import csv
 from io import StringIO
--- a/l10n_br_fiscal/__manifest__.py
+++ b/l10n_br_fiscal/__manifest__.py
@@ -133,6 +133,7 @@
     "installable": True,
     "application": True,
     "auto_install": False,
+    "post_init_hook": "post_init_hook",
     "external_dependencies": {
         "python": [
             "erpbrasil.base",
--- a/l10n_br_fiscal/hooks.py
+++ b/l10n_br_fiscal/hooks.py
@@ -1,0 +1,33 @@
+# Copyright 2025-TODAY Akretion - Raphael Valyi <raphael.valyi@akretion.com>
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0.en.html).
+
+from odoo import SUPERUSER_ID, api
+
+
+def fill_missing_xml_ids(env):
+    """
+    Backfill missing ir.model.data (XML IDs) for manually created records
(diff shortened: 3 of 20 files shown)


antoniospneto:
@rvalyi 

Ainda não olhei a PR com atenção, mas pretendo revisar, apenas uma observação por enquanto:

Nesses casos a tag pra por no commit é [IMP] ou [REF] 
[ADD] é somente quando é módulo novo:
https://www.odoo.com/documentation/16.0/contributing/development/git_guidelines.html#tag-and-module-name

rvalyi:
nota: ver se não seria possivel usar o mixin no tax.definition. O @rodmad85 teve o caso e tb tive o caso semana passada ao migrar um banco. Teria que ver se tem como bolar um xml_id único no caso do tax.definition...

rvalyi:
vou validar isso num banco de prod esse fds, mas dei uma boa limpada, completei para os registros de tax.pis.cofins e de tax.definition e deve estar pronto ou quase.

rvalyi:
@OCA/local-brazil-maintainers Pessoal, agora que apareceu mais um caso de atualizacao de dados aqui https://github.com/OCA/l10n-brazil/pull/4776 (e teve do PIS/COFINS uns dias atras tb), achei bom dar um rebase e uma limpada neste PR meu para resolver esses caso de forma mais sistematica, visto que eh uma problematica bem importante no Brasil.

kaynnan (approved):

kaynnan (approved):

PR #4416, opened 2026-03-05
[16.0] l10n_br_fiscal_dfe, l10n_br_nfe_dfe: generic l10n_br_fiscal_dfe, add l10n_br_nfe_dfe for NF-e DF-e

rvalyi:
o problema de #4383 é que ele torna o modulo l10n_br_fiscal_dfe especifico para a NF-e.

Mas poderia ser útil usar serviço de DF-e para outros documentos fiscais como a CT-e (alias na Akretion vamos precisar para um cliente). Por isso eu acho melhor conservar um modulo l10n_br_fiscal_dfe genérico.

E ai em vez de entupir o modulo l10n_br_nfe que já ta mega pesado de coisas extra de DF-e com dependências novas (tipo queue_job), eu prefiro seguir o que faz o sucesso da OCA: módulos pequenos e fáceis de migrar (ou ainda pequenos para caber no contexto de uma IA...). Dessa forma, se "alguém" preferir um modulo de DFe que usa erpbrasil.edoc terminado no m*jo por estagiários de faculdade, sem problema desde que não vem atrasar a vida de geral na OCA. E ai pelo menos temos um l10n_br_nfe robusto e fácil de migrar.

EDIT: nessa outra branch (para depois), eu fiz um prototipo de como poderia ser o modulo de DF-e especifico para CT-e:

- https://github.com/akretion/l10n-brazil/tree/16.0-generic-dfe-cte/l10n_br_cte_dfe
- ele poderia usar esse cliente de DF-e de CT-e inspirado no cliente da NF-e: https://github.com/akretion/nfelib/pull/148

Disclaimer, acabei de fazer o refator com AI mas eu ainda não testei em produção.

PR #4465, opened 2026-03-27
[18.0][PERF] l10n_br_fiscal_dfe: fix N+1 query in _records_to_attachments

marcos-mendez:
## Problem

`_records_to_attachments()` issued one `SELECT` per record to collect
`ir.attachment` rows:

```python
for record in attachment_ids:
    ids += attachment_obj.search([('res_id', '=', record.id)])
```

With N fiscal documents selected for batch download, this produced N
sequential round-trips to PostgreSQL. On a typical monthly closing with
200+ documents the overhead is measurable and load scales linearly with
batch size.

Additionally, the original query filtered only by `res_id` without
`res_model`, meaning it could inadvertently return attachments from
unrelated models that share the same integer ID.

## Fix

Replace the loop with a single batched query:

```python
attachment_ids = attachment_obj.search([
    ('res_model', '=', attachment_ids._name),
    ('res_id', 'in', attachment_ids.ids),
])
```

## Impact

| | Before | After |
|---|---|---|
| Queries | N (one per record) | 1 |
| `res_model` filter | missing | added |

## Testing

Tested with `l10n_br_nfe` MDE batch download selecting multiple
records — behavior unchanged, single query confirmed via PostgreSQL logs.

### DIFF
--- a/l10n_br_fiscal_dfe/models/attachment.py
+++ b/l10n_br_fiscal_dfe/models/attachment.py
@@ -100,9 +100,11 @@ class Attachment(models.TransientModel):
             attachment_ids = attachs
 
         if attachment_ids._name != "ir.attachment":
-            ids = attachment_obj
-            for record in attachment_ids:
-                ids += attachment_obj.search([("res_id", "=", record.id)])
-            attachment_ids = ids
+            attachment_ids = attachment_obj.search(
+                [
+                    ("res_model", "=", attachment_ids._name),
+                    ("res_id", "in", attachment_ids.ids),
+                ]
+            )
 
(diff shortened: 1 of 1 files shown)


rvalyi (changes requested):
Este modulo é uma bosta que nunca deveria ter entrado assim na OCA... (so ver quem costumava fazer isso...)
Existe um refactor extenso em curso #4416

marcos-mendez:
A OK, so to passando o que fiz nos ultimos 6 meses na industria porque essa parte tava lenta. Quer que feche a PR?

PR #4506, opened 2026-04-25
[WIP][16.0] l10n_br_sped_efd_icms_ipi

mileo:


rvalyi (changes requested):
@mileo if you had search for the addition PR for this module you would have find my open PR https://github.com/OCA/l10n-brazil/pull/4043 already...

So instead of adding random AI slope for us to review (like we don't have enough work clearly), you might have just asked on the PR why the PR was not merged or what what the status instead of creating a new PR...

Well since you didn't ask, here is the answer anyway: the reason it was not merged, is because of the broken mapping KMEE already contributed previously under your direction sadly... And also because we were too busy fixing important bugs or migration related changes while KMEE don't even maintain the modules you forced into the repo over the years...

The OCA has never been about "fake it until you make it", in more than 13 years you are around that something you should have ended up learning I think... The OCA would never exist along Odoo SA if it behaves in the same superficial manner... But also the whole idea of the OCA is that people can trust OCA code instead of contracting an infinite technical debt with code that is obviously not even tested...
The problem is 30% of the mappings you put in the l10n_br_sped_efd_icms_ipi where completely broken, like not even close to running and the signal to noise is even worse in this new PR...

If you had done a PSC job of following the PRs (or is it only for linkedIn auto-congratulations?), at least the one with work from KMEE, you would have noticed this report from @crsilveira for instance on v14 yet: https://github.com/OCA/l10n-brazil/pull/3104#issuecomment-3019890142

But I also had to fix many of your other mappings before cherry-picking them and other more recently completely broken like https://github.com/OCA/l10n-brazil/pull/4043/changes/ee7d80d2902882be32d603f66bdffb57f1379616
So basically I did the layout update in a way that is possible to review and in a way that leave the code up to date (unless your random chars changes for quotes in your PR)  https://github.com/OCA/l10n-brazil/pull/4043

So I cherry-picked and updated your mappings for C100 and D100 that were correct and the fix for "C100: Fix _odoo_domain to filter NF-e document types"
here https://github.com/OCA/l10n-brazil/pull/4043/changes/d66937a9f0e9b708d3b36b8c7ea4978b68dacc2f
Inside this PR, your mappings proposals for C190 and D190 are completely broken as these are grouping registers, not register that should be mapped to each single fiscal document lines! 

You E110 mapping is also yet another gross joke... These kinds of tax analysis registers are quite touchy, and I think we should 1st merge the l10n_br_sped_efd_icms_ipi module (once it's free from random shitty mappings) and review these kind of mappings very carefully...

But also the problem is that we should adapt the l10n_br_sped_efd_icms_ipi layout to the layout 20 (mandatory since January) and sadly it was quite hard to update. I'll post about the process later, but you can get an idea with my update of the spedextractor here https://github.com/akretion/sped-extractor/pull/13

So closing this one for now. BTW, we are about to go for homologation for #4043 and likely this Monday we will set it read for review as well.

cc @renatonlima @marcelsavegnago @antoniospneto @DiegoParadeda @AndreMarcos @kaynnan @CristianoMafraJunior

PR #4513, opened 2026-04-25
[18.0][FW] [17.0][FIX] l10n_br_mdfe_spec: avoid test conflict

rvalyi:
Port from 17.0 to 18.0:\n- Commit 04923479


rvalyi:
NOTE: we might do that only once l10n_br_mdfe is ported if we have conflict issues like in the 17.0 branch. We should not do port that now because it would kill the test coverage.

PR #4528, opened 2026-05-03
[16.0][REF] l10n_br_account: mv tests to move_workflow

rvalyi:
move some l10n_br_account tests to a dedicated test_move_workflow.py file (test_account_move_lc.py is only for account move entry tests)


CristianoMafraJunior (approved):
LGTM

marcelsavegnago (approved):
LGTM

PR #4655, opened 2026-07-09
[16.0][ADD] l10n_br_duimp: add new module

kaynnan:
cc @marcelsavegnago @WesleyOliveira98 @CristianoMafraJunior 

## Summary

Adds a new module, `l10n_br_duimp`, that queries a DUIMP (Declaração Única de Importação) directly from the Portal Único Siscomex REST API and generates the corresponding inbound fiscal document and vendor bill in Odoo — instead of requiring manual data entry from the customs broker's extract.

While implementing and testing this end-to-end (query → fiscal document → invoice), two pre-existing bugs in `l10n_br_fiscal` and `l10n_br_account` were found and fixed, since they blocked the flow entirely. Both are included as separate commits in this PR (see "What was necessary to change" below); the new module does not work without them (verified by reverting each fix and re-running the test suite, which fails at the exact points described).

### `l10n_br_fiscal` — fix malformed OR domain in partner search

`document.import.wizard._search_partner()` appended the legal_name/name OR-domain as a nested list instead of extending the domain:

domain.append(["|", ("legal_name", "ilike", legal_name), ("name", "ilike", name)])

This produces an invalid leaf and raises `ValueError` whenever a partner is looked up by name only (no CNPJ) — exactly the DUIMP case, since a foreign exporter has no CNPJ to match on.

### `l10n_br_account` — fix `import_fiscal_document()` for company-issued documents

Two issues in the same method, only reachable when actually running the full "create fiscal document → generate invoice" flow end-to-end:

- `move._compute_fiscal_amount()` doesn't exist: `account.move` only gets `l10n_br_fiscal.document` fields via `_inherits` delegation, not its methods. Fixed to call it on `fiscal_document_id`.
- `document_serie` (Char) was written unconditionally, but it's invisible/read-only in the form whenever the fiscal document `issuer` is the company itself (e.g. a self-issued import entry note) — `document_serie_id` (the internal serie catalog) must be used instead. Fixed to branch on `issuer`, matching the view's own `attrs`.

## What was identified (design notes / limitations)

- The DUIMP has no official downloadable XML like NFe/CTe/MDFe — it is only available via the authenticated REST API (mTLS with the company's e-CNPJ digital certificate). The wizard queries by number/version instead of parsing an uploaded file.
- ICMS is not returned by this DUIMP service (it's a state, not federal, tax) — its base/value fields stay for manual entry.
- AFRMM is not returned by this DUIMP service either (it's handled by Siscomex Carga/Mercante) — entered manually on the wizard and allocated to lines proportionally to customs value.
- II/IPI/PIS/COFINS totals come from the DUIMP's header-level `tributos.tributosCalculados` and are allocated to each line proportionally to its `valorAduaneiro` (customs value); freight and insurance are used directly from the DUIMP's own per-item allocation (`valorFreteRateado` / `valorSeguroRateado`).
- Tax base/amount fields remain freely editable after import, reusing the existing `imported_document` mechanism (same one already used for NFe/CTe/MDFe XML import) — no new override mechanism was introduced.
- The exact JSON field names were validated against the official Siscomex docs and a community OpenAPI-generated client (github.com/samuelfac/portalunico.siscomex.gov.br), not against a live payload. `_duimp_exporter_name()` in particular is flagged as best-effort and should be adjusted against a real response before production use.

## Testing

- New automated test (`test_duimp_import_wizard.py`) mocks the Siscomex webservice and covers: querying, matching an item to a product/CFOP, importing, and confirming the imported tax values survive a save without being recomputed by the standard fiscal tax engine.
- Not yet tested against a real Siscomex environment (homologação/produção) with a live e-CNPJ certificate.


rvalyi (changes requested):
Uma coisa que eu não acho muito legal é ter o "account" no nome o módulo. pois de fato esse módulo mexe muito mais na parte fiscal e faz quase nada com account... ai o nome poderia ser apenas l10n_br_duimp Talvez (a parte fiscal ficando meio implicita). Seria tb interessante o @renatonlima revisar porque temos clientes fazendo a DI faz uns 12 anos de uma forma ou outra...

mileo:
@kaynnan pode dar uma rebase?

mileo:
@kaynnan pode revisar os testes e depois voltar para open?

kaynnan:
@mileo Feito.

PR #4711, opened 2026-07-17
[16.0][IMP] l10n_br_fiscal, l10n_br_account: performance optimizations for the invoice/sale fiscal flow

mileo:
Code-only performance optimizations for the fiscal document flow. The performance test framework that measured these was split into a separate opt-in PR so this one can be reviewed as pure code.

Four low-risk patches:

1. `l10n_br_fiscal`: short-circuit empty recordset in `map_tax_definition`. Avoids up to 3 guaranteed-empty searches per line when company/CFOP/fiscal-profile have no tax definitions (common in production).
2. `l10n_br_fiscal`: skip redundant `fiscal_tax_ids` re-sync in `_update_fiscal_tax_ids`. Only reassigns when the set actually changes, avoiding a second tax-engine pass on persisted writes.
3. `l10n_br_account`: drop global `clear_caches()` from move/aml unlink. `_sync_dynamic_lines` deletes and recreates term/tax lines on every save, so these unlinks wiped all worker ormcaches (views, ACLs, translations) inside the create/post flow itself. The ORM already invalidates deleted records, and git archaeology found no recorded motivation for the global clear. Systemic win: on production with N workers the effect is multiplicative.
4. `l10n_br_account`: guard `update_payment_term_number` writes. Filters by `fiscal_operation_id` and `is_invoice`, and only writes when the value changes.

Measured impact (30-line invoice, Form path): post -40%, invoice_post -52%, so_create -67% queries. The full before/after numbers are in the split framework PR.

Functional suites of l10n_br_fiscal / l10n_br_account / l10n_br_sale: 0 failed. The 13 pre-existing errors in l10n_br_sale (a CacheMiss on account.move.journal_id) are present on pristine oca/16.0 too and are unrelated to this change.

PR #4716, opened 2026-07-17
[16.0][IMP] l10n_br_fiscal, l10n_br_account: per-transaction fiscal mapping/compute_taxes memoization (perf wave 2)

mileo:
## Summary

Wave 2 of the fiscal-engine performance work: **per-transaction memoization of the fiscal tax mapping and of `compute_taxes`**, plus batching of the `account_taxes` mapping.

Builds on #4711 (now merged into 16.0). Rebased on top of it, so this PR is standalone and its diff is only the wave-2 code. The performance test framework is a separate opt-in PR (#4741); the query numbers below were measured with it.

## Motivation

Even after #4711, the invoice Form path costs ~250 queries/line: the fiscal mapping does several searches per line (`map_tax_definition` x4, `icms_regulation.map_tax` x4, `line_definition`, `account_taxes` = 2 per tax), and `compute_taxes` is re-executed 3-6x per line with identical inputs within the same save.

## Commits

- `[IMP] l10n_br_account`: `account_taxes` mapping in **one search per tax group** instead of 2 searches per tax per line.
- `[IMP] l10n_br_fiscal`: **per-transaction memoization of the fiscal mapping** + fuse the 4 `icms_regulation.map_tax` searches into 1.
- `[IMP] l10n_br_fiscal`: **memoize `compute_taxes` per transaction**.
- `[ADD] l10n_br_fiscal`: regression test for the compute_taxes cache across a savepoint rollback.

## Cache design (`l10n_br_fiscal/models/fiscal_cache.py`)

- Anchored on the cursor (`env.cr._l10n_br_fiscal_txn_cache`): one cursor = one transaction; every request/job starts empty; never a module-level dict (no leaking across transactions/workers).
- Keys include all mapping inputs (record ids + branch-driving scalars: tax_framework, company tax classification, states, `ind_ie_dest`, `icms_origin`, operation type...). Unexpected kwarg type -> memoization disabled for that call (never risks a wrong hit).
- Invalidation: `FiscalCacheMixin` inherited by `tax.definition`, `icms.regulation`, `tax.classification`, `operation.line` and `tax` - any create/write/unlink clears the current transaction's cache.
- `compute_taxes` returns a **copy** of the cached result: callers mutate the dict (e.g. `account_tax.py` flips the base sign for refunds), which would otherwise corrupt subsequent hits.

## ICMS search fusion

The 4 searches (ICMS, ICMS-ST, FCP/DIFAL, FCP-ST) become one search over the OR of the 4 domains. Each sub-domain fixes a distinct `tax_group_id`, so the result partitions cleanly per group and the original precedence (tax benefit > specific > generic, then `ind_final`) runs per partition, extracted to `_tax_definition_precedence`. The `_map_tax_def_*` sub-methods are kept (they may be overridden downstream), just no longer called internally.

## Measured impact (query counts, warm round, minimal DB)

| Step | Before | After | delta |
|---|---|---|---|
| Form create 30 lines | 7572 | **5874** | **-22.4%** |
| Form create 2 lines | 732 | 616 | -15.8% |
| so_create 30 lines | 360 | **120** | **-67%** |
| so_create 2 lines | 116 | 88 | -24% |
| Form post 2 / 30 | 70 / 78 | 67 / 75 | -4% |

Note on methodology: query counts are deterministic and reproducible (the numbers above hold on any machine); CI/runtime time varies too much between runners to prove a figure from a single run, so the guard asserts query counts and a scaling ratio, not time.

Functional suites of `l10n_br_fiscal`, `l10n_br_account` and `l10n_br_sale`: **0 failed**.

### DIFF
--- a/l10n_br_account/models/fiscal_tax.py
+++ b/l10n_br_account/models/fiscal_tax.py
@@ -7,10 +7,57 @@ from odoo import Command, api, fields, models
 class FiscalTax(models.Model):
     _inherit = "l10n_br_fiscal.tax"
 
+    def _account_taxes_company(self, company=False):
+        if company:
+            return company
+        # ``self.env.company`` already resolves ``allowed_company_ids[0]`` *with*
+        # access validation, so it replaces the previous raw browse of that id.
+        # A ``default_company_id`` in the context is still honoured as an
+        # explicit override.
+        company = self.env.company
+        if self.env.context.get("default_company_id"):
+            company = self.env["res.company"].browse(
+                self.env.context["default_company_id"]
+            )
+        return company
+
     def account_taxes(self, user_type="sale", fiscal_operation=False, company=False):
         account_taxes = self.env["account.tax"]
+        if not self:
+            return account_taxes
+
+        company = self._account_taxes_company(company)
+
+        # Resolve os grupos contábeis de todos os impostos fiscais em um único
+        # search e traz TODOS os account.tax dos grupos envolvidos numa única
+        # query (antes: 1 search de grupo + 1 search de account.tax por imposto
+        # por linha). O casamento por grupo é feito em memória.
+        fiscal_group_to_account_group = self.tax_group_id._account_tax_group_map()
+        account_group_ids = [
+            group.id for group in fiscal_group_to_account_group.values() if group
+        ]
+        candidate_taxes = self.env["account.tax"].search(
+            [
+                ("tax_group_id", "in", account_group_ids),
+                ("active", "=", True),
+                ("company_id", "=", company.id),
+            ]
+        )
+        taxes_by_account_group = {}
+        for tax in candidate_taxes:
+            taxes_by_account_group.setdefault(
+                tax.tax_group_id.id, self.env["account.tax"]
+            )
+            taxes_by_account_group[tax.tax_group_id.id] |= tax
+
         for fiscal_tax in self:
-            taxes = fiscal_tax._account_taxes(company)
+            account_group = fiscal_group_to_account_group.get(
+                fiscal_tax.tax_group_id.id
+            )
+            taxes = taxes_by_account_group.get(
+                account_group.id if account_group else False,
+                self.env["account.tax"],
+            )
             # Atualiza os impostos contábeis relacionados aos impostos fiscais
             account_taxes |= taxes.filtered(
                 lambda t: t.type_tax_use == user_type and t.active and not t.deductible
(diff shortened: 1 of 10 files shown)


mileo:
Rebaseado sobre o #4711 (mergeado): PR agora standalone, so o codigo da Onda 2. Historico limpo - fixups da review dobrados nos commits e prefixo [16.0] removido dos commits (fica so no titulo do PR). Proximo da fila incremental. Suite funcional verde; numeros de query na descricao.

mileo:
## Field evidence: profiled on a real production database

Measured on a restored production database (a real one, not demo data: on the order
of 100 thousand invoices and 200 thousand fiscal document lines, with the full set of
localization modules installed).

Creating and posting the same 10-line invoice, profiled in-process with cProfile on
the same database and with the same input payload, only the code differing. The first
invoice of the process is discarded, so the trigger tree construction is not counted.

Reporting **call counts and query counts only**: they are properties of the code path,
independent of the machine that runs them, which is what makes them comparable across
environments.

| tree | Python calls |
|---|---|
| `oca/16.0` | 15,549,073 |
| `oca/16.0` + this series (#4716 -> #4717 -> #4727 -> #4721) | **7,095,071** |
| delta | **-8,454,002 (-54%)** |

Where those 8.45 M calls went, all of them ORM record traversal primitives:

| function | `oca/16.0` | with the series | delta |
|---|---|---|---|
| `builtins.len` | 1,452,624 | 218,453 | -1,234,171 |
| `models.py:__iter__` | 1,112,718 | 131,052 | -981,666 |
| `models.py:__init__` (recordset instantiation) | 1,017,401 | 277,992 | -739,409 |
| `fields.py:convert_to_record` | 599,641 | 106,481 | -493,160 |
| `fields.py:mapped` / `models.py:mapped` | 275,612 / 275,365 | 30,056 / 29,809 | -245,556 each |
| `api.py:get_until_miss` (cache lookup) | 274,532 | 28,976 | -245,556 |

That is the signature of repeatedly walking the same recordsets: instantiate, iterate,
map a field, convert a cached value back into a record. The expected counterpart shows
up too and is far smaller: slightly more SQL execution and one more `_flush_search`.

Query count per invoice on the same database: **1,566 to 1,126 (-28%)**, that is 806 to
555 on create and 760 to 571 on post.

Note on method: we deliberately do not quote wall-clock deltas. Besides depending on
the machine, our RPC harness builds the `create` payload from each tree's own onchange
output, so the two runs would not receive identical input. Counts do not have either
problem.

kaynnan (approved):

PR #4719, opened 2026-07-18
[16.0][FIX] l10n_br_nfse: other_retentions_value never injected (dead fields_view_get override)

mileo:
## Summary

`l10n_br_nfse` overrides `fields_view_get` on `l10n_br_fiscal.document.line` to inject `other_retentions_value` into the fiscal view - but on Odoo 16 that override is **dead code**: the modern view pipeline (`get_views` -> `_get_view`) never calls the model-level `fields_view_get` (it only survives as a reverse-compatibility wrapper in `ir.ui.view`; there are no callers in core nor in l10n-brazil). On top of that, the injection body was wrapped in a bare `except Exception` that silently swallowed any failure.

Net effect: **`other_retentions_value` has not been present in the fiscal line views since the 16.0 migration**, and nothing ever reported it.

## Fix

Migrate the injection to the `_get_view` hook (same mechanism used by the fiscal view injection everywhere else in 16.0) and drop the silent exception handling.

Verified empirically: before the patch the field is absent from the combined arch returned by the modern pipeline; after the patch it is present again. `l10n_br_nfse` test suite green.

Found while working on the fiscal view performance series (#4711/#4716/#4717), but this is an independent functional bug fix - submitted standalone so it can merge fast.

### DIFF
--- a/l10n_br_nfse/models/document_line.py
+++ b/l10n_br_nfse/models/document_line.py
@@ -1,10 +1,14 @@
 # Copyright 2020 KMEE INFORMATICA LTDA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+import logging
+
 from erpbrasil.base import misc
 from lxml import etree
 
 from odoo import api, fields, models
+
+_logger = logging.getLogger(__name__)
 
 
 class DocumentLine(models.Model):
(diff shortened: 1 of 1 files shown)


CristianoMafraJunior:
@mileo esta em draft?

mileo:

marcelsavegnago:
@mileo acha que consegue dar uma melhorada na cobertura de teste ?

PR #4720, opened 2026-07-18
[16.0][IMP] l10n_br_fiscal: extensible fiscal view field hooks + prune dead injected fields

mileo:
## Summary

Two related improvements to the fiscal line view injection (`inject_fiscal_fields`), with no behavior change for standard flows:

1. **Extensibility hooks** - the injection contract becomes overridable downstream:
 - `_fiscal_view_compute_methods()`: formalizes the (previously hardcoded) triad of compute methods whose fields are collected into line views;
 - `_fiscal_view_pruned_fields()`: the list of fields excluded from injection, so custom modules can re-add a pruned field (or prune more) with a simple `super()` override.

2. **Prune 5 dead injected fields** - `ii_percent`, `simple_value`, `simple_without_icms_value`, `cofins_wh_base_type`, `pis_wh_base_type`. Census across the whole repo (views, modifiers/attrs/domains, `@api.onchange` triggers, compute bodies, JS, QWeb): these are stored computed fields injected only through the synthetic `missing_line_fields` append, never read by any onchange, never referenced by any modifier, never displayed. They keep being computed and stored server-side - they only leave the onchange payload/view spec. Value-preserving by construction.

Fields that had any uncertain usage (e.g. `amount_tax_included`, which IS read by `_sync_invoice`; the `*_cst_code` related fields read by nfse/downstream) were deliberately **kept** - the rule applied was "any doubt -> keep".

The injection tests in `l10n_br_purchase` and `l10n_br_sale_blanket_order` were extended to assert the new contract (they are the executable spec of the injection).

Part of the fiscal performance series (#4711/#4716/#4717) as groundwork for view-payload work, but standalone and independent - this PR is deliberately perf-neutral.

### DIFF
--- a/l10n_br_fiscal/models/document_line_mixin.py
+++ b/l10n_br_fiscal/models/document_line_mixin.py
@@ -103,6 +103,46 @@ class FiscalDocumentLineMixin(models.AbstractModel):
         return domain
 
     @api.model
+    def _fiscal_view_compute_methods(self):
+        """Compute methods whose (stored) fields must be kept available in the
+        business-document line views (invisible), so the fiscal onchange
+        protocol can round-trip them. Formalizes the implicit triad used by
+        ``inject_fiscal_fields``. A submodule adding a fiscal compute to the
+        line can extend it::
+
+            @api.model
+            def _fiscal_view_compute_methods(self):
+                return super()._fiscal_view_compute_methods() + [
+                    "_compute_my_tax_fields",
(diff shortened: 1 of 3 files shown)


mileo:

marcelsavegnago:
@mileo ficou em conflito

mileo:
> @mileo ficou em conflito

resolvido

PR #4721, opened 2026-07-18
[16.0][IMP] l10n_br_fiscal: reduce injected tree field collection (-33% Form time)

mileo:
> Note: the performance test framework was split into a separate opt-in PR (#4741). This PR contains only the code change, reviewable on its own.

## What it does

Reduces the fiscal field collection injected into the **business-document line TREES** (`account.move`, `sale.order`, `purchase.order`, stock, blanket orders) from ~240 to ~120 fields, keeping the **full fiscal detail editable in the line form dialog** (the popup opened from the tree - the web client loads it as an isolated datapoint, outside the line onchange spec).

Why the tree: with the default editable tree the o2m **edition view is the tree**, and the per-line onchange payload/spec is built from the tree fields only (`basic_model.js` uses `fieldsInfo["list"]`; `odoo.tests.Form` resolves the same way). Every extra invisible field in that collection is fetched/computed on every onchange round, for every line - that is where the read-side cost of the Form path lives.

Whitelist kept in the tree (`_fiscal_line_view_fields`): the `FISCAL_TAX_ID` tax selectors, every `@api.onchange` trigger and every stored **non-computed** field. The modifier/domain/context closure of the injected nodes is force-kept (`_fiscal_line_modifier_refs`), so a reduced tree can never raise *"Unknown field ... in modifier"* at view load. The ~121 removed fields are all stored **computed** fields - they keep being computed and stored server-side, so dropping them from the inline onchange collection is value-preserving.

## Prerequisite - stacked on #4727

This PR is **stacked on #4727** (`[16.0][FIX] draft tax override clobbered by spurious compute re-fire`), which is itself stacked on the perf series #4711 -> #4716 -> #4719 -> #4720. Only the **top commit** is new here; the perf baseline tests from #4711 live in the branch so CI measures this patch objectively.

#4727 is a hard prerequisite: it decouples draft fiscal correctness from the view payload (deterministic server-side recompute). Without it, reducing the onchange spec regressed the draft ICMS rate; stacked on it, the value is correct. Full mechanism of that regression in the **root-cause comment**: https://github.com/OCA/l10n-brazil/pull/4721#issuecomment-5011764003

## Measured (perf baseline from #4711, warm round, `perf-oca`)

| Step | Before (on #4727) | After |
|---|---|---|
| create_form 30 lines | 5874q / ~20.6s | **5571q / ~13.7s (-33% wall time)** |
| create_form 2 lines | 616q / ~1.29s | 599q / ~1.07s |
| post / so_create / create_invoices | unchanged | unchanged |

Query count is deterministic; wall time is logged, never asserted (machine-dependent). Reproduce: check out the branch and run `--test-tags l10n_br_performance` on the l10n_br_account/l10n_br_sale suites.

## Test migration - tree column -> line popup

Removing the ~121 computed fields from the tree also removes them as **optional tree columns**, which a few tests edited/read directly through the inline `Form`. Those were migrated to preserve their intent (the fiscal detail is now edited in the line dialog, not the tree):

- `l10n_br_account` `TestMoveEdition.test_out_fiscal_invoice`: the manual FCP override (`icmsfcp_base` / `icmsfcp_value`) is now set through the line form dialog (popup) `Form` and asserted to persist through save.
- `l10n_br_account` `TestPerTaxDraftEdition`: the IPI/PIS/COFINS override cases select the tax live (the selector stays in the tree) and assert the per-tax value on the **saved fiscal line** - the exact value #4721 could leave stale. ICMS base/value/percent stay in the tree, so ICMS is still checked live on the inline Form.

Suites `l10n_br_fiscal` / `l10n_br_account` / `l10n_br_sale`: **0 failed** (incl. `test_out_fiscal_invoice` and the full `TestPerTaxDraftEdition`).

## Structural note

The residual cost of the 30-line Form is protected by the compute web (the stored fiscal totals/related recompute on every line touch). Reducing it further is an Onda-3 structural change (C5/RFC), not a view prune - this PR takes the safe payload reduction that #4727 makes correct.

### DIFF
--- a/l10n_br_account/models/fiscal_tax.py
+++ b/l10n_br_account/models/fiscal_tax.py
@@ -7,10 +7,65 @@ from odoo import Command, api, fields, models
 class FiscalTax(models.Model):
     _inherit = "l10n_br_fiscal.tax"
 
+    def _account_taxes_company(self, company=False):
+        if company:
+            return company
+        # ``self.env.company`` already resolves ``allowed_company_ids[0]`` *with*
+        # access validation, so it replaces the previous raw browse of that id.
+        # A ``default_company_id`` in the context is still honoured as an
+        # explicit override.
+        company = self.env.company
+        if self.env.context.get("default_company_id"):
+            company = self.env["res.company"].browse(
+                self.env.context["default_company_id"]
+            )
+        return company
+
     def account_taxes(self, user_type="sale", fiscal_operation=False, company=False):
         account_taxes = self.env["account.tax"]
+        if not self:
+            return account_taxes
+
+        company = self._account_taxes_company(company)
+
+        # Resolve os grupos contábeis de todos os impostos fiscais em um único
+        # search e traz TODOS os account.tax dos grupos envolvidos numa única
+        # query (antes: 1 search de grupo + 1 search de account.tax por imposto
+        # por linha). O casamento por grupo é feito em memória.
+        fiscal_group_to_account_group = self.tax_group_id._account_tax_group_map()
+        account_group_ids = [
+            group.id for group in fiscal_group_to_account_group.values() if group
+        ]
+        candidate_taxes = self.env["account.tax"].search(
+            [
+                ("tax_group_id", "in", account_group_ids),
+                ("active", "=", True),
+                ("company_id", "=", company.id),
+            ]
+        )
+        taxes_by_account_group = {}
+        for tax in candidate_taxes:
+            taxes_by_account_group.setdefault(
+                tax.tax_group_id.id, self.env["account.tax"]
+            )
+            taxes_by_account_group[tax.tax_group_id.id] |= tax
+
+        # `deductible_taxes` is company dependent, so reading it off the record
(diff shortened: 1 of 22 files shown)


mileo:
**Root cause found - and fixed in #4727.** The regression was not a missing input in the reduced spec: the user's `icms_tax_id` override was being **clobbered by a spurious compute re-fire** (shared multi-field compute + foreign-write remarking; full mechanism in the #4722 comment and in #4727's description).

With #4727 underneath, this branch passes `test_out_fiscal_invoice` **and** the new per-tax draft value suite, with the measured gain intact: **5.874q/20,6s -> 5.571q/~13,7s (-33% wall time)** on the 30-line Form benchmark.

Next: this PR will be rebased on top of #4727 and the remaining test references to tree-only optional columns (e.g. `icmsfcp_base`) migrated to the line dialog, then marked ready for review.

volkkov (commented):
Logic looks correct. One suggestion: the early return on line prevents the cleanup function from running — worth adding a finally block.

mileo:

PR #4727, opened 2026-07-18
[16.0][FIX] l10n_br_fiscal: draft tax override clobbered by spurious compute re-fire (+ per-tax value tests)

mileo:
> Note: the performance test framework was split into a separate opt-in PR (#4741). This PR contains only the code change, reviewable on its own.

## Summary

Fixes a **latent fiscal-correctness fragility that exists in 16.0 today**, found while investigating the #4721 regression, and adds the **per-tax draft edition value test suite** promised in RFC #4722 (§6).

**Stacked on the perf series (#4711 -> #4716 -> #4717, + #4719/#4720 commits). Only the last 2 commits are new here.**

## The bug class this fixes (root cause of the #4721 regression - but present without it)

When a user manually overrides a tax on a draft line (e.g. sets `icms_tax_id` from the 7% to the 18% tax), that override can be **silently clobbered** back to the mapped tax by a spurious compute re-fire. Two vectors, both confirmed by instrumentation:

- **Vector A - shared multi-field compute**: `_compute_fiscal_tax_ids` computed 5 fields together (`fiscal_tax_ids`, `cfop_id`, `ipi_guideline_id`, `tax_classification_id`, `icms_tax_benefit_id`, all `store` + `precompute` + `readonly=False`). Odoo gotcha: manually assigning ONE output of a multi-field editable compute does **not** protect the sibling outputs - if any sibling is still marked to-compute, reading it re-runs the whole compute, which re-maps `fiscal_tax_ids` from the operation line and discards the user's choice.
- **Vector B - foreign-write remarking**: `_compute_product_fiscal_fields` re-assigned product-derived inputs with **unchanged** values; each such write triggers `modified()`, re-marking `fiscal_tax_ids` as dirty, with the same clobbering effect.

In the current full view spec these re-fires are mostly **accidentally suppressed** (all fields travel in the onchange payload and sit in cache), which is why the bug is latent - and why reducing the view spec (#4721) exposed it as a wrong ICMS amount. Correctness should not depend on which fields a view happens to carry.

## Fix (surgical, no schema change, no new depends fan)

1. **Split the multi-field compute**: `_compute_fiscal_tax_ids` now computes only `fiscal_tax_ids`; the operation-line decoration fields move to a new `_compute_fiscal_operation_data`. `fiscal_tax_ids` only recomputes when its true inputs change - never as a side effect of reading a decoration field. (`map_fiscal_taxes` is memoized since #4716, so the extra compute call is effectively free.)
2. **Delta-write in `_compute_product_fiscal_fields`** (same pattern as #4717): only fields whose value actually changes are written, so redundant re-assignments no longer re-mark the fiscal mapping.

An idempotent "reconcile mapped vs. overridden taxes" approach was evaluated and rejected: there is no local signal that distinguishes a user override from a stale derived value (tracking override intent would require new state - structural, 18.0 material; see RFC #4722).

## Per-tax draft edition value tests (`test_per_tax_draft_edition.py`)

For ICMS, IPI, PIS and COFINS: change the line's `*_tax_id` in a draft Form and assert base/amount follow the NEW rate; change quantity/price_unit; switch `fiscal_operation_line_id` and assert CFOP/taxes remap. Green on the current baseline; these tests are the safety net the RFC declares as prerequisite for any structural work (ISSQN scenario omitted: the LC demo setup sells goods - documented in the file).

## Measured (perf baseline from #4711)

Query-neutral on the Form benchmark (the remaining cost there is read-side protocol prefetch), ~17% wall-time improvement locally from eliminated re-computes. The real payoff: **it unlocks #4721** - with this fix underneath, the tree-collection reduction passes all value tests (-5.2% queries / ~-33% time measured, with correct fiscal values). #4721 will be rebased on top of this PR.

Functional suites `l10n_br_fiscal` / `l10n_br_account` / `l10n_br_sale`: 0 failed. Performance guards from #4711: green, limits untouched.

### DIFF
--- a/l10n_br_account/models/fiscal_tax.py
+++ b/l10n_br_account/models/fiscal_tax.py
@@ -7,10 +7,65 @@ from odoo import Command, api, fields, models
 class FiscalTax(models.Model):
     _inherit = "l10n_br_fiscal.tax"
 
+    def _account_taxes_company(self, company=False):
+        if company:
+            return company
+        # ``self.env.company`` already resolves ``allowed_company_ids[0]`` *with*
+        # access validation, so it replaces the previous raw browse of that id.
+        # A ``default_company_id`` in the context is still honoured as an
+        # explicit override.
+        company = self.env.company
+        if self.env.context.get("default_company_id"):
+            company = self.env["res.company"].browse(
+                self.env.context["default_company_id"]
+            )
+        return company
+
     def account_taxes(self, user_type="sale", fiscal_operation=False, company=False):
         account_taxes = self.env["account.tax"]
+        if not self:
+            return account_taxes
+
+        company = self._account_taxes_company(company)
+
+        # Resolve os grupos contábeis de todos os impostos fiscais em um único
+        # search e traz TODOS os account.tax dos grupos envolvidos numa única
+        # query (antes: 1 search de grupo + 1 search de account.tax por imposto
+        # por linha). O casamento por grupo é feito em memória.
+        fiscal_group_to_account_group = self.tax_group_id._account_tax_group_map()
+        account_group_ids = [
+            group.id for group in fiscal_group_to_account_group.values() if group
+        ]
+        candidate_taxes = self.env["account.tax"].search(
+            [
+                ("tax_group_id", "in", account_group_ids),
+                ("active", "=", True),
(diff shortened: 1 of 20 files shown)


mileo:

PR #4789, opened 2026-07-27
[16.0][FIX][IMP] l10n_br_fiscal, l10n_br_account, l10n_br_nfe: import follow-ups

mileo:
Small follow-ups on the fiscal document import layer, after #4494 and #4653.
One atomic commit per change:

## [FIX] l10n_br_account: allow importing a 100% free line

`_check_document_import` blocked any line with a zero unit price, which
rejects legitimate free lines (bonificacao / free sample declared with
vUnCom=0) faithfully imported from the XML. The zero price is now only
flagged when the line still carries amounts, which is an actually
unresolved de-para. Tests cover the free line, the inconsistent
zero-price line and the missing product case.

## [FIX] l10n_br_fiscal: scope import wizard line ACL to fiscal users

The generic import wizard line was granted full access to every internal
user (`base.group_user`). It is now scoped to `l10n_br_fiscal.group_user`,
like the other fiscal wizards.

## [IMP] l10n_br_nfe: log the ICMS tax created during import

The import can create an ICMS fiscal tax with base reduction when the
company has none configured (#4653). Master data created as an import
side effect now leaves a log trail so the fiscal responsible can review
it afterwards.

## [IMP] l10n_br_fiscal: stamp date_in_out on the imported document

The effective incoming/outgoing date (DT_E_S in the SPED bookkeeping)
was never filled by the import flow. The import wizard now exposes it,
defaulting to the import date, and stamps it on the imported document
when it is not set yet. This is one of the fields a correct fiscal
bookkeeping of purchases needs (C100 registry).

## Tests

Full suites of l10n_br_fiscal, l10n_br_account, l10n_br_nfe and
l10n_br_account_nfe pass locally on Odoo 16 (232 tests, 0 failed,
0 errors, nfelib 2.5.2).


antoniospneto (approved):
LGTM

rvalyi (approved):

CristianoMafraJunior (approved):

PR #4806, opened 2026-07-29
[16.0][FIX] l10n_br_mdfe_spec: index the one2many inverse foreign keys

mileo:
## Problem

The generated spec models declare no index on the many2one fields that serve as
the inverse key of a one2many (`xsd_implicit=True` + `ondelete="cascade"`).

In Odoo 16 a `Many2one` does not create an index by default, and PostgreSQL does
not index foreign key columns automatically. Reading such a one2many issues
`SELECT ... WHERE <inverse_fk> = X` against an unindexed column, which is a
sequential scan of the whole child table.

`l10n_br_mdfe_spec` has 34 one2many and, before this PR, **0** fields with `index=True`.
This PR indexes 34 of them.

## Measured on a production database

Restored production database of a Brazilian customer (82,214 fiscal documents,
587 new documents per month, PostgreSQL 15):

| relation | rows | plan | time |
|---|---|---|---|
| `res.partner.nfe40_autXML_infNFe_id` | 73,359 | Seq Scan | **106 ms** |
| the same after this change | 73,359 | Index Scan | **0.064 ms** |
| for reference, `l10n_br_fiscal.document.line.document_id` (already indexed) | 204,698 | Index Scan | 1.3 ms |

## Honest scope

I also measured the end-to-end invoice flow (create + post) on that database,
before and after adding these indexes: **no significant change** (2.339s -> 2.337s,
median of 3 runs). The test invoice does not populate those groups, so the scans
do not run in that path.

The gain shows up where many documents are read at once: batch XML export,
reports, DF-e import, and any `search()` on those relations. So this is a
correctness/hygiene fix with a clear worst case, **not an invoice speed-up** -
flagging it explicitly so it is not oversold.

## Why applied to the generated files instead of regenerating

The change mirrors exactly what the generator emits after
akretion/xsdata-odoo#22 (`index=True` in `odoo_implicit_many2ones`).

I did try a full regeneration first, and verified by comparing field definitions
semantically (not textually) that it would be **destructive** with the current
toolchain:

- the output is not even valid Python (`from __future__ import annotationsfrom
  .tipos_basico_v4_00 import (`);
- nfelib 2.5.2 does not carry the tax reform types, so 6 IBS/CBS and IS fields
  would degrade from `Many2one`/`Selection` to `Char`, **losing** support that is
  already committed here;
- 74 fields would disappear.

So this PR touches only the implicit many2one fields and nothing else. The diff is
`index=True` plus the line rewrapping it causes. Whoever regenerates the specs
next with the fixed generator will produce the same result.

## Verification

- the number of modified fields matches the number of one2many in the module;
- every file still compiles;
- pre-commit (ruff, ruff-format, pylint_odoo) passes.

### DIFF
--- a/l10n_br_mdfe_spec/models/v3_0/mdfe_modal_aquaviario_v3_00.py
+++ b/l10n_br_mdfe_spec/models/v3_0/mdfe_modal_aquaviario_v3_00.py
@@ -141,7 +141,10 @@ class InfTermCarreg(models.AbstractModel):
     _binding_type = "Aquav.InfTermCarreg"
 
     mdfe30_infTermCarreg_aquav_id = fields.Many2one(
-        comodel_name="mdfe.30.aquav", xsd_implicit=True, ondelete="cascade"
+        comodel_name="mdfe.30.aquav",
+        xsd_implicit=True,
+        ondelete="cascade",
+        index=True,
     )
     mdfe30_cTermCarreg = fields.Char(
         string="Código do Terminal de Carregamento",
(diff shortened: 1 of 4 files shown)


rvalyi (approved):

CristianoMafraJunior (approved):
LGTM

PR #4809, opened 2026-07-29
[16.0][FIX] l10n_br_fiscal: tax engine must not swallow compute errors

mileo:
### 1. `compute_taxes` silently replaces a crashed computation with the generic formula

`l10n_br_fiscal/models/tax.py` wrapped both the `getattr` lookup and the call to
the per-domain compute method in the same `try`:

```python
try:
    compute_method = getattr(self, f"_compute_{tax.tax_domain}")
    taxes[tax.tax_domain].update(compute_method(tax, taxes, **kwargs))
except AttributeError:
    taxes[tax.tax_domain].update(tax._compute_tax(tax, taxes, **kwargs))
```

The intent is a fallback for tax domains that have no specialized method. But the
`except` also catches any `AttributeError` raised INSIDE `_compute_icms`,
`_compute_icmssn`, etc. (a `None` dereference, a typo, a field renamed in a
migration). When that happens the engine posts a plausible but wrong tax value
computed by the generic formula, with nothing in the log.

Fixed with `getattr(self, ..., None)`: the fallback behavior for domains without
a specialized method is unchanged, and programming errors now propagate.

### 2. The accounting entry point cannot provide `icms_cst_id`, and FCP crashes on it

`_compute_icmsfcp` did:

```python
icms_cst_id = kwargs.get("icms_cst_id")   # None when absent
...
if icms_cst_id.code in ICSM_CST_CSOSN_ST_BASE:   # AttributeError on None
```

`account.tax.compute_all` (l10n_br_account) has no `icms_cst_id` parameter in its
signature, so every call coming from the accounting side reaches this line with
`None` and raises. Combined with defect 1, the crash was invisible: on the
accounting side FCP silently fell back to the generic formula, while the fiscal
document side (which passes `icms_cst_id`) used the specialized one. Same line,
two different computations, depending on the entry point.

Fixed by defaulting to an empty recordset, the exact pattern `_compute_icms`
already uses:

```python
cst = kwargs.get("icms_cst_id", self.env["l10n_br_fiscal.cst"])
```

## The tests

`test_tax_engine_contract.py` pins the engine contract itself, on purpose without
relying on fiscal configuration or demo data (records are resolved by `search`,
and the compute methods are called directly with a well-formed tax dictionary):

- `test_icmsfcp_without_icms_cst_id`: the accounting entry point must be able to
  compute FCP;
- `test_icmsfcp_with_and_without_icms_cst_id_agree`: the two entry points must
  agree on the same line;
- `test_compute_taxes_does_not_hide_programming_errors`: an `AttributeError`
  raised inside a compute method must propagate (asserted by injecting one).

All three fail on current 16.0 and pass with this PR.

## Behavior change to be aware of

On the accounting side, FCP for CST/CSOSN in `ICSM_CST_CSOSN_ST_BASE` is now
computed by the specialized method instead of the silent generic fallback. That
is the intended correction, but databases that relied on the previous accounting
values will see them change. Flagging it for reviewers rather than hiding it.

## How this was found

Reviewing the engine after a performance investigation on a restored production
database: the `except AttributeError` showed up as the reason a crash in the FCP
path had never surfaced. Full module test suite runs with no new failures
relative to the same database without this change.

### DIFF
--- a/l10n_br_fiscal/models/tax.py
+++ b/l10n_br_fiscal/models/tax.py
@@ -561,7 +561,7 @@ class Tax(models.Model):
         tax_dict = taxes_dict.get(tax.tax_domain)
         partner = kwargs.get("partner")
         company = kwargs.get("company")
-        icms_cst_id = kwargs.get("icms_cst_id")
+        icms_cst_id = kwargs.get("icms_cst_id") or self.env["l10n_br_fiscal.cst"]
 
         if taxes_dict.get("icms"):
             if company.state_id != partner.state_id:
@@ -835,11 +835,14 @@ class Tax(models.Model):
             operation_line = kwargs.get("operation_line")
             fiscal_operation_type = operation_line.fiscal_operation_type or FISCAL_OUT
             kwargs.update({"cst": tax.cst_from_tax(fiscal_operation_type)})
-            try:
-                compute_method = getattr(self, f"_compute_{tax.tax_domain}")
+            # getattr with a default instead of try/except AttributeError: the
+            # previous form also caught AttributeError raised INSIDE the compute
+            # method, silently replacing a specialized computation that crashed
+            # with the generic formula.
+            compute_method = getattr(self, f"_compute_{tax.tax_domain}", None)
+            if compute_method is not None:
                 taxes[tax.tax_domain].update(compute_method(tax, taxes, **kwargs))
-
-            except AttributeError:
+            else:
                 taxes[tax.tax_domain].update(tax._compute_tax(tax, taxes, **kwargs))
 
             tax_domain = taxes[tax.tax_domain]
--- a/l10n_br_fiscal/tests/__init__.py
+++ b/l10n_br_fiscal/tests/__init__.py
@@ -1,6 +1,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from . import (
+    test_tax_engine_contract,
     test_cnae,
     test_fiscal_document_serie,
     test_fiscal_document_generic,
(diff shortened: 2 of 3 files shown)


mileo:
Fechando em favor do #4838, que ataca o mesmo problema e ja carrega a cobertura
daqui.

Os dois PRs mexem no mesmo ponto: o `except AttributeError` em volta do dispatch de
`compute_taxes`, em `l10n_br_fiscal/models/tax.py`, que troca em silencio o metodo
especifico pelo calculo generico e deixa o imposto sair errado sem erro nenhum. O #4838
resolve com 34 linhas (o dispatch passa a resolver o metodo com um default em vez de
capturar o erro em volta da chamada) contra 8 aqui, e veio do cruzamento com a anomalia
do #4692, entao ficou com o diagnostico mais completo.

O que este PR tinha de proprio era o contrato do FCP, e isso nao se perde: as duas
assercoes foram para o #4838, em `test_fiscal_tax.py`. Sao elas que o
`account.tax.compute_all` exige, porque ele nao tem `icms_cst_id` na assinatura:

- `test_icmsfcp_does_not_require_icms_cst_id`: o FCP tem de ser calculavel sem o
  parametro, senao a entrada contabil levanta AttributeError dentro do motor;
- `test_icmsfcp_agrees_with_and_without_icms_cst_id`: o valor tem de ser o mesmo com e
  sem ele, senao o mesmo documento sai com FCP diferente conforme quem o precificou.

Uma observacao que vale para quem escrever teste parecido: elas foram para dentro do
`test_fiscal_tax.py` e nao para um arquivo proprio de proposito. Um modulo de teste que
faz `from ..models.tax import ...` e e carregado pelo `tests/__init__.py` fecha um ciclo
de import, e a carga do registry falha com `cannot import name ... (most likely due to a
circular import)`. Os testes do modulo que importam do proprio addon ou nao estao no
`__init__.py` (o `test_ibpt.py`) ou importam de `constants`, nao de `models`.

O trabalho segue no #4838.

PR #4880, opened 2026-08-09
[16.0][IMP] spec_driven_model: disambiguate dup labels

rvalyi:
[FIX] spec_driven_model: disambiguate duplicate spec field labels

    When NFe, CTe, MDFe... XSD spec fields are injected into the same Odoo
    model (res.partner, l10n_br_fiscal.document...), many of them share the
    same label coming from the XSD schema, making Odoo emit a
    'Two fields ... have the same label' warning for each such pair of
    fields. These warnings are not legit in the spec_driven_model context:
    the labels come from the XSD schemas and cannot be changed without
    diverging from the spec.

    So we now dynamically disambiguate the labels of the spec fields that
    would be duplicate by appending their schema and XSD tag, e.g. 'Nome do
    município (cte_xMun)' or 'Dados do endereço (cte_enderExped)'. Only
    spec fields involved in a duplicate get disambiguated, regular fields
    keep their labels untouched.

    The disambiguation runs in SpecModel._setup_fields for the spec models
    proper, and right before the duplicate label check of
    ir.model.fields._reflect_fields for any model receiving spec fields
    through another mechanism (typically the _inherits delegation used by
    l10n_br_account for account.move, account.payment or
    account.bank.statement.line), because the label of a delegated field is
    copied from its parent model during the child setup, possibly before
    the parent _setup_fields disambiguation, depending on the model setup
    order. _reflect_fields runs after all model setup, whatever the way the
    spec fields were injected.

**antes**: mais de 1300 WARNING como odoo odoo.addons.base.models.ir_model: Two fields (cte40_toma, service_provider) of l10n_br_fiscal.document() have the same label: Tomador do Serviço. [Modules: l10n_br_cte and l10n_br_fiscal] 
**depois**: 11 casos de warning de same labels que nao tem a ver com os modulos que usam spec_driven_model

remover esses warnings eh necessario para mesclar a migraçao dos modulos l10n_br_mdfe e l10n_br_cte na branch 18.0 onde warnings nao sao mais permitidos.

### DIFF
--- a/l10n_br_cte/models/document.py
+++ b/l10n_br_cte/models/document.py
@@ -212,8 +212,8 @@ class CTe(spec_models.StackedModel):
                 and record.document_type_id.prefix
                 and record.document_key
             ):
-                record.cte40_Id = "{}{}".format(
-                    record.document_type_id.prefix, record.document_key
+                record.cte40_Id = (
+                    f"{record.document_type_id.prefix}{record.document_key}"
                 )
 
     ##########################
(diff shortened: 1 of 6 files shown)


mileo (approved):

CristianoMafraJunior (approved):
LGTM

PR #4920, opened 2026-08-11
[17.0][FW] l10n_br_nfe: multiple ports from 16.0

rvalyi:
Port from 16.0 to 17.0:
- #4494
- #4653
- #4681
- #4692
- #4762
- #4789


CristianoMafraJunior (approved):

mileo (approved):

rvalyi:
@mileo please take a look at the PRs and diff before merging anything blindly (we can click on the button too, it doesn't help you do it), the point is these changes are important and require database update so it should be merge minor at the minimum. patch is only when no database update is required (only a restart).

But in fact changes such as https://github.com/OCA/l10n-brazil/pull/4653 are quite important and were merged as major, so the port should propagate this information.

/ocabot merge major

PR #4921, opened 2026-08-11
[18.0][FW] l10n_br_nfe: multiple ports from 16.0

rvalyi:
Port from 16.0 to 18.0:
- #4494
- #4653
- #4681
- #4692
- #4789

These PRs are ported for all modules already migrated to 18.0 (l10n_br_fiscal, l10n_br_account, l10n_br_nfe). Sub-commits belonging to l10n_br_cte and l10n_br_account_nfe are NOT ported (those modules are not available on 18.0 yet).

PR #4762 (Lucro Real company demo data) is not included in this port — it requires the br_oca_generic chart template setup on 18.0 which is a separate stream.

### DIFF
--- a/l10n_br_account/data/template/account.tax.group-br_oca.csv
+++ b/l10n_br_account/data/template/account.tax.group-br_oca.csv
@@ -22,3 +22,5 @@
 "tax_group_inss_wh","l10n_br_fiscal.tax_group_inss_wh"
 "tax_group_simples","l10n_br_fiscal.tax_group_simples"
 "tax_group_others","l10n_br_fiscal.tax_group_others"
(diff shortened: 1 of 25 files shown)


CristianoMafraJunior (approved):

mileo (approved):

rvalyi:
@mileo please take a look at the PRs and diff before merging anything blindly (we can click on the button too, it doesn't help you do it), the point is these changes are important and require database update so it should be merge minor at the minimum. patch is only when no database update is required (only a restart).

But in fact changes such as https://github.com/OCA/l10n-brazil/pull/4653 are quite important and were merged as major, so the port should propagate this information.

/ocabot merge major

PR #4925, opened 2026-08-12
[16.0][FIX] spec_driven_model: import the binding module instead of assuming it is imported

mileo:
_get_binding_class resolves the binding module with sys.modules[], which
assumes something already imported it. NF-e, CT-e and MDF-e import their
bindings at the top of their files so they never hit this, but a module
with an optional external dependency cannot import it at load time (an
import warning at loading fails the checklog) and gets a KeyError on the
first export. Found while developing l10n_br_reinf.

The fix resolves the module with importlib.import_module, which returns the
cached module when it is already imported (no behavior change for the
existing documents).

### DIFF
--- a/spec_driven_model/models/spec_export.py
+++ b/spec_driven_model/models/spec_export.py
@@ -3,7 +3,7 @@
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0.en.html).
 
 import logging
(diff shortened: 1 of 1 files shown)


rvalyi (approved):
LGTM

CristianoMafraJunior (approved):
LGTM

PR #4791, opened 2026-07-27
[16.0][IMP] l10n_br_fiscal: per-line import review state and persisted de-para (draft-first import, 1/3)

mileo:
First of 3 stacked PRs evolving the vendor XML import flow, building on the
foundation merged in #4494 and #4653 (thanks @rvalyi!). Depends on #4789
(small follow-ups, stacked below).

## The problem

Today the de-para review (product matching, UoM conversion, CFOP remapping)
lives in the TransientModel wizard and is applied by rewriting the parsed
binding (`_edit_parsed_xml`) before the fiscal document is built. This works
for the one-shot manual flow, but:

1. A transient review cannot survive a page refresh nor be shared between
   users: a 200-line NF-e with a half-done de-para loses everything.
2. There is no pending queue. An unmatched product line has nowhere to wait
   for a human decision (we want product creation to always be a supervised
   action, never an import side effect).
3. Flows with no user on the screen (DF-e capture from SEFAZ, batch upload
   of a zip with several XMLs) need a persisted destination for the review.
4. Rewriting the binding loses the audit trail of the supplier nomenclature
   and already needs workarounds (e.g. the manual price_unit preservation
   in `_create_edoc_from_file`).

Brazilian ERPs converge on the same answer: a persisted "pre-document"
holding the review state (Protheus pre-nota, Sankhya import portal). This
is a clean 16.0 reimplementation of the ideas of the old KMEE PR #3546,
converging with the generic wizard line of #4653.

## What this PR does (foundation only, no flow change yet)

- `l10n_br_fiscal.document.line` gains a review state (pending / matched /
  resolved), the de-para fields (internal UoM, conversion factor, supplier
  info link) and a snapshot of the supplier nomenclature
  (`partner_uom_code` / `partner_quantity` / `partner_price_unit`,
  following the `partner_cfop_id` pattern of #4653), plus a CFOP alert
  computed against the real issuer/company geography (states, countries,
  foreign trade CFOPs).
- `_apply_import_depara()` is the single write path of the de-para:
  snapshot, unit conversion preserving the line total (quantity times
  factor, unit price divided by it), fiscal operation, supplier info
  learning. Re-applying converts from the original snapshot, so a wrong
  factor can be fixed without compounding.
- `_suggest_product()` (through supplier nomenclature hooks each document
  type implements) and `_suggest_fiscal_operation_in()` through the
  INVERSE CFOP: the supplier declares its outbound CFOP (5xxx/6xxx), the
[...]

PR #4938, opened 2026-08-13
[16.0][IMP] spec_driven_model: context-driven record creation policy for XML imports 

rvalyi:
### Size breakdown (for reviewers)

The PR shows +458 / -245 lines, but the framework code is actually more
compact. The extra lines are the new features and the new tests:

- ~205 lines of pure boilerplate **removed**: the per-model
  `match_or_create_m2o` overrides that only prevented creation, and the
  per-comodel search-keys files (see "Removed boilerplate" below).
- ~170 lines of **new tests** (create policy matrix in spec_driven_model,
  import diagnostics in l10n_br_nfe).
- ~160 lines for the **new fine-grained create policy** feature (context
  keys, import wizard checkbox, diagnostic error messages).
- ~90 lines for the **new schema-level m2o search keys** mechanism
  (declarations on the schema mixins + generic reading in spec_import).
- The remaining ~-30 lines are net simplifications in the module models.

### What / Why  (AI assisted feature)

The spec_driven_model import framework currently auto-creates any record it
can't match (`match_or_create_m2o`), which forces every consumer module
(l10n_br_nfe, l10n_br_cte, l10n_br_mdfe...) to override `match_or_create_m2o`
on each model just to prevent creation. Those overrides are copy-paste
boilerplate, and there's no way to opt in/out per import scenario (supplier
NFe vs go-live loading of the company's own NFe).

This PR adds a generic, context-driven policy so callers can
blacklist/whitelist which models may be created, and decide what happens when
a forbidden record has no match.

### New context keys (all optional, fully backward compatible)

- `spec_create_forbidden_models` — list of model names never to create, or a
  dict {model: 'skip'|'raise'} for per-model behaviour.
- `spec_create_allowed_models` — if set, only these models may be created
  (whitelist).
- `spec_create_forbidden_action` — default action for the list form: 'skip'
  (leave the m2o empty, historical behaviour) or 'raise'.

### Product creation

Importing an NFe that can't map a product previously silently created a
partial product. The import wizard now has a "Create unmapped products"
checkbox (default: disabled). When disabled, an unmapped product raises a
clear UserError stating the model, the searched keys/domain, and the XML
values that failed to match.

### Removed boilerplate

The `match_or_create_m2o` overrides that only did "search then return False"
were removed across l10n_br_nfe (res.country, res.country.state, res.city,
uom, cest, tax.ipi.guideline), l10n_br_cte (res.country, res.country.state,
res.city) and l10n_br_mdfe (res.country.state). Matching is done by the
generic `match_record` and creation is prevented by the context key.

The per-comodel search-keys override files are gone as well: l10n_br_nfe's
res_city.py, res_country_state.py, cest.py and tax_ipi_guideline.py,
l10n_br_cte's res_city.py, res_country.py and res_country_state.py, and
l10n_br_mdfe's res_city.py and res_country_state.py. Their
`_{schema}_search_keys` / `_{schema}_extra_domain` declarations now live on
the schema mixin (see next section).

Overrides that do real business matching (res.partner, product.product) now
respect the policy by delegating to spec.mixin before creating.

### M2o search keys on the schema mixin

`match_record` / `_spec_import_search_info` now read additional per-comodel
search keys (and extra domains) from the schema mixin
(`spec.mixin.<schema>`, e.g. `spec.mixin.nfe` in l10n_br_nfe_spec) through a
schema-prefixed attribute (`_nfe_m2o_search_keys` / `_nfe_m2o_extra_domain`),
with fallback to the comodel `_{schema}_search_keys` / `_{schema}_extra_domain`
attributes as before. Each `*_spec` module declares the keys its schema
needs — e.g. res.city matched by ibge_code, res.country.state by
ibge_code+code, l10n_br_fiscal.cest / tax.ipi.guideline by code_unmasked for
NFe. This removes the last per-comodel files while keeping the matching
schema-scoped and generic.

Also fixed the "FIXME make generic": `match_record` now reads
`_{schema}_extra_domain` instead of hardcoding `_nfe_extra_domain`, so NFe and
CTe both work.

### Tests

- spec_driven_model: blacklist prevents create, blacklist still matches
  existing, whitelist restricts, raise action (with message content
  assertions), backward-compat default, registry reload with extended
  remaining model, hook read-only on ordinary registry load.
- l10n_br_nfe: product creation forbidden raises with diagnostic message;
  imported document workflow test adapted to the creation policy.

OBS: esse PR casa bem com esse outro https://github.com/OCA/l10n-brazil/pull/5076


mileo:
Nice, this is the right layer for the creation policy, and it converges with
what the draft-first import series (#4791, #4793, #4795) is building on top of
the same wizard: there the unmapped product does not block the import, the
line is persisted as pending with the supplier data snapshotted, and a review
queue + per-line resolution dialog handle the mapping afterwards (supervised
creation included). In other words: your policy's 'skip' action is exactly the
entry point of that flow, and 'raise' stays perfect for the go-live scenario.

Two convergence points we would like to align on:
1. We will adapt our stacked work to consume this policy from the context
   (spec_create_forbidden_models={'product.product': 'skip'}) instead of
   introducing a parallel flag, and keep your allow_product_creation field as
   the single knob on the wizard.
2. Merge order: this PR first, then #4939, then the draft-first series
   rebased on top (and #4940 after, since it touches the same wizard files).

Does that work for you?

rvalyi:
vem mais uma simplificaçao ainda...

PR #4946, opened 2026-08-14
[16.0][FIX] l10n_br_mdfe: match cities by ibge_code on import

rodmad85:
Fixes the MDF-e import on environments where `l10n_br_nfe` / `l10n_br_cte` are not installed.

Without this, `res.city` has no `match_or_create_m2o` override and the MDF-e import
tries to create a city with empty values, raising a `NotNullViolation` on
`country_id`. That breaks the `post_init_hook` sample import and the
`test_import_in_mdfe` test on a minimal (deps-only) OCA CI install, blocking
any PR touching `l10n_br_mdfe` (e.g. #4860).

Mirrors the existing behavior of `l10n_br_nfe`/`l10n_br_cte`: match by
`ibge_code` and never create a city on import.

Tested with a CI-faithful run (`oca_init_test_database` + `oca_run_tests`):
0 failed, 0 error(s) of 31 tests.

### DIFF
--- a/l10n_br_mdfe/models/__init__.py
+++ b/l10n_br_mdfe/models/__init__.py
@@ -1,6 +1,7 @@
 from . import transporte
 from . import res_company
 from . import res_country_state
+from . import res_city
 from . import document
 from . import document_related
 from . import res_partner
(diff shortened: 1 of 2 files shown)


mileo (approved):

rvalyi (approved):
Repara que uma parte desse boiler plate vai ficar inútil porém depois de mesclar https://github.com/OCA/l10n-brazil/pull/4938
Mas pode entrar assim por enquanto

PR #5136, opened 2026-09-11
[18.0][FIX] l10n_br_coa: load classification from chart CSV

marcelsavegnago:
## Summary
- The `18.0.2.1.0` post-migration searched `account.account.template`, which was removed in Odoo 17 and is not in the 18.0 registry.
- That `KeyError` aborted `-u l10n_br_coa` on databases already at `18.0.2.0.0`.
- The script now reads `tag_ids` / `tag_ids:id` from the chart CSV and copies missing classification tags onto existing accounts via `{company_id}_{xmlid}`.

## Test plan
- [x] Upgrade a database with `l10n_br_coa` `18.0.2.0.0` (`-u l10n_br_coa`) and confirm the post-migration no longer raises `KeyError: 'account.account.template'`.
- [x] On a company that already loaded `br_oca_generic` or `br_oca_simple`, confirm accounts receive the missing `l10n_br_coa.account_tag_*` tags.
- [x] Confirm accounts that were tagged by hand keep their existing tags (migration only links missing ones).
- [x] Confirm companies without a Brazilian OCA chart are left unchanged.

Verified on `develbrcoa`:
- Upgrade completed without `KeyError`.
- `Empresa Lucro Real` + `br_oca_generic`: migration tagged **392/493** accounts.
- After clearing tags on 15 accounts, a second run restored them.
- A manual tag (`Manual Classification Test`) was kept; template tags were only linked.
- `generic_coa` companies (`Sua Empresa`, `Lucro Presumido`, `Simples Nacional`) stayed at **0** Brazilian classification tags.

PR #5153, opened 2026-09-14
[17.0][FIX] l10n_br_coa: load classification from chart CSV

marcelsavegnago:
account.account.template was removed in Odoo 17, so the 17.0.1.2.0 post-migration crashed with KeyError and blocked module upgrades. Read tag_ids from the chart template CSV instead and copy missing tags onto already loaded accounts.

port from: #5136


rvalyi (approved):

rvalyi:
Already merged in 18.0

/ocabot merge patch

CristianoMafraJunior (approved):

PR #5160, opened 2026-09-15
[16.0][FIX] l10n_br_sale_blanket_order: open standalone line form

MarcusPDL:
Opening an existing record from Sales → Orders → Blanket Order Lines raises an OwlError caused by `Cannot read properties of undefined (reading 'fiscal_operation_id')`. The standalone line form evaluates modifiers against `parent`, which is only available in a relational subview.

Use the line's own fiscal operation for fiscal field visibility, and its partner, pricelist and company for the fiscal quantity context and account tax domain. Include the existing related `pricelist_id` as an invisible field so it is available to the client.

### Validation

- XML parsing and inheritance XPath targets checked against `sale_blanket_order` 16.0.
- Isolated evaluation reproduced the missing-parent failure before the fix; 24 fiscal visibility cases passed after the fix (six fields, ICMS/ISSQN, with/without an operation), along with quantity contexts and company tax domains.
- Prettier 2.7.1 with XML plugin 2.2.0 and `git diff --check` passed.
- Full Odoo/browser integration has not been run locally.

### Manual verification

1. Open an existing line through Sales → Orders → Blanket Order Lines.
2. Check fiscal fields for goods and services, and with no fiscal operation on the line.
3. Check opening the line from the blanket order, the fiscal quantity and the Accounting tab tax selection.


marcelsavegnago:
@CristianoMafraJunior @MarcusPDL Coloquei a PR como rascunho.. Quando de fato finalizar ai libera para revisão

